PR #65, opened 2016-08-20
fix Lintian documentation/man issues

satta:
This PR addresses some issues I encountering while packaging:

 - spelling: "dont" -> "don't"
 - `lambda_indexer` didn't have a short description, leaving the corresponding field empty in the manpage (`--export-help man`)
 - I had to add an extra word in front of `'45'` as a singlequote at the beginning of a line apparently refers to a macro in the groff format, which leads to an error in the manpage (`--export-help man`)


h-2:
thanks for the PR :+1:

PR #121, opened 2019-01-14
Enable and update tests for lambda2

mr-c:
Only the first 10 tests pass with this patch


h-2:
Did you add the checksums that you observed locally, or did you take the checksums from the previously gzipped files?

h-2:
superseded by #159

PR #130, opened 2019-07-18
Lambda3 main

h-2:
commits will be stashed. 

@sarahet Do you want to have a look at this?

 We don't have CI, yet.

### DIFF
--- a/src/lambda.cpp
+++ b/src/lambda.cpp
@@ -19,8 +19,7 @@
 // lambda.cpp: Main File for Lambda
 // ==========================================================================
 
-#include <seqan/seqan3_compat.h>
-#include <seqan/seqan3_compat_post.h>
+#include "seqan2seqan3.hpp"
 
 #include <seqan3/argument_parser/all.hpp>
 
--- a/src/lambda.cpp
+++ b/src/lambda.cpp
@@ -24,10 +24,7 @@
 #include <seqan3/argument_parser/all.hpp>
 
 #include "search.hpp"
-
-#if 1
 #include "mkindex.hpp"
-#endif
 // using namespace seqan;
 
 void parseCommandLineMain(int argc, char const ** argv);
--- a/src/lambda.cpp
+++ b/src/lambda.cpp
@@ -19,6 +19,9 @@
 // lambda.cpp: Main File for Lambda
 // ==========================================================================
 
+#include <seqan/seqan3_compat.h>
+#include <seqan/seqan3_compat_post.h>
+
 #include <seqan3/argument_parser/all.hpp>
 
 #include "search.hpp"
@@ -88,7 +91,7 @@ int main(int argc, char const ** argv)
 
     if ((std::string(argv[until]) == "mkindexp") || (std::string(argv[until]) == "mkindexn"))
     {
-    #if 0
+    #if 1
         try
         {
             mkindexMain(argc - until, argv + until);
--- a/src/lambda.cpp
+++ b/src/lambda.cpp
@@ -21,11 +21,11 @@
 
 #include <seqan3/argument_parser/all.hpp>
 
-#if 0
 #include "search.hpp"
-#endif
-#include "mkindex.hpp"
 
+#if 1
+#include "mkindex.hpp"
+#endif
 // using namespace seqan;
 
 void parseCommandLineMain(int argc, char const ** argv);
@@ -71,7 +71,7 @@ int main(int argc, char const ** argv)
     // TODO change return values
     if ((std::string(argv[until]) == "searchp") || (std::string(argv[until]) == "searchn"))
     {
-#if 0
+
         try
         {
             searchMain(argc - until, argv + until);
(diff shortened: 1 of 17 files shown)


src/lambda.cpp
```diff
@@ -19,13 +19,15 @@
...
+
 #include <seqan3/argument_parser/all.hpp>
 
-#if 0
 #include "search.hpp"
-#endif
-#include "mkindex.hpp"
 
+#if 1
+#include "mkindex.hpp"
```
sarahet: Does the search now always get included? And why did that change? I guess the macro was initially meant to distinguish between search and index

src/mkindex.hpp
```diff
@@ -202,10 +192,7 @@ void realMain(LambdaIndexerOptions     const & options)
     // see if final sequence set actually fits into index
 //     checkIndexSize(translatedSeqs, options, seqan::BlastProgramSelector<p>());
 
-    std::tie(f.redSeqs, f.index) =
-        generateIndex<c_dbIndexType == DbIndexType::BI_FM_INDEX, c_redAlph>(f.transSeqs, options);
-
-
+    f.index = generateIndex<c_dbIndexType == DbIndexType::BI_FM_INDEX>(f.transSeqs, options);
```
sarahet: This means we always have a bidirectional index?

src/mkindex_algo.hpp
```diff
@@ -236,7 +234,13 @@ translateSeqs(TCDStringSet<std::vector<TOrigAlph>> & in,
...
     myPrint(options, 1, "Translating Subj Sequences...");
 
-    out = in | seqan3::view::translate | std::view::join; //TODO geneticCode
+    auto tmp  = in | seqan3::view::translate //TODO geneticCode
+                   | std::view::join
+                   | std::view::transform([] (auto && elem)
+                     {
+                         return std::forward<decltype(elem)>(elem) | std::ranges::to<std::vector>;
+                     })
+                   | std::ranges::to<decltype(out)>;
```
sarahet: Are the two views after ```view_join``` used in order to make it comparable to later seqan2 types?

src/mkindex_algo.hpp
```diff
@@ -125,7 +123,7 @@ loadSubjSeqsAndIds(LambdaIndexerOptions const & options)
             extractAccIds(id, count);
 
         if (options.truncateIDs)
-            ids.push_back(id | seqan3::view::take_until(seqan3::is_space));
+            ids.push_back(id | seqan3::view::take_until(seqan3::is_space) | std::ranges::to<std::string>);
```
sarahet: Why?

src/mkindex.hpp
```diff
@@ -52,15 +52,9 @@ template <DbIndexType   c_indexType,
           AlphabetEnum  c_origAlph>
```
sarahet: ```#pragma once``` missing in this file?

src/mkindex_options.hpp
```diff
@@ -57,12 +55,12 @@ struct LambdaIndexerOptions : public SharedOptions
 // INDEXER
 void parseCommandLine(LambdaIndexerOptions & options, int argc, char const ** argv)
 {
-    std::string programName = "lambda3 " + std::string(argv[0]);
+    std::string programName = "lambda3-" + std::string(argv[0]);
```
sarahet: what is ```lambda3-```? :smile:

src/search_algo.hpp
```diff
@@ -1354,686 +661,6 @@ _writeRecord(TBlastRecord & record,
 // Function computeBlastMatch()
 // --------------------------------------------------------------------------
 
-template <typename TBlastRecord,
```
sarahet: Why don't we need this huge block anymore?

src/lambda.cpp
h-2: It's a leftover from debugging!

src/mkindex.hpp
h-2: no, it's a `==` not a `=`

src/mkindex_algo.hpp
h-2: No, they convert from view to range (which is quite cumbersome for two-dimensional ranges).

src/mkindex_algo.hpp
h-2: Views are no longer implicitly convertible to containers, you need to pipe them into `ranges::to` to convert.

src/mkindex.hpp
h-2: Good catch!

src/mkindex_options.hpp
h-2: program names are no longer allowed to contain whitespace in SeqAn3 🤷‍♂

src/search_algo.hpp
h-2: It's the old "hand-crafted" back-tracking algorithm. it is replaced by standard search now.

Later-on is the "hand-crafted" seed-and-extend approach which I have gotten rid of now since it won't ever work with SeqAn3 and is so heuristic, it's almost random 😉

src/mkindex.hpp
sarahet: Oh right

src/search_algo.hpp
sarahet: I remember that we talked about it :smile:

h-2:
mac is broken, but we can fix it another time

PR #133, opened 2019-10-31
Fixes

h-2:


h-2:
The second commit only fixes the broken results with BlastN. The general search questions are not yet answered.

(but its much easier to debug blastn than any translated mode)

src/search_algo.hpp
```diff
@@ -367,7 +367,26 @@ loadQuery(GlobalDataHolder<c_indexType, c_origSbjAlph, c_transAlph, c_redAlph, c
         throw QueryException{"Zero sequences submitted."};
     }
 
-    // TODO if constexpr BLASTN mode, reverse complements
+    if constexpr (TGH::blastProgram == seqan::BlastProgram::BLASTN)
+    {
+        //TODO view-ify this at some point. Better yet, move to index.
```
sarahet: Why move to index at some point?

src/search_algo.hpp
```diff
@@ -367,7 +367,26 @@ loadQuery(GlobalDataHolder<c_indexType, c_origSbjAlph, c_transAlph, c_redAlph, c
...
+        // ATTENTION, this won't work well if qryseqs is concatenated_sequences
```
sarahet: What do we do in this case? Is there a better way then?

src/search_algo.hpp
h-2: Well, copying and revcomplementing doubles the number of query sequences which doubles the run-time of the program (linear influence).

 Reverse-complementing instead the database sequences only has logarithmic influence on the run-time.
OTOH it doubles the required memory which is also very significant.

It's a difficult call.

src/search_algo.hpp
h-2: In that case it would be better to work on the underlying vector directly, because one can pre-allocate the entire thing with one operation.

But this will go away if we stream over the input sequences so I don't think we should worry about it now.

PR #138, opened 2019-11-14
Update lambda3 from master

sarahet:
Finally :sparkles:

### DIFF
--- a/INFO
+++ b/INFO
@@ -6,7 +6,7 @@ Copyright: 2013-2019, Hannes Hauswedell; 2016-2019 Knut Reinert, FU-Berlin
 Status: under development
 Description: Lambda is a biological sequence aligner optimized for many
 query sequences and searches in protein space. It is highly compatible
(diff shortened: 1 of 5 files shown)


h-2:
Ok! And the macOS build on lambda3 is broken anyway, right?

sarahet:
> Ok! And the macOS build on lambda3 is broken anyway, right?

I think so, at least we ignored it for your last PR as well. But maybe we should find out why at some point ..

h-2:
> But maybe we should find out why at some point ..

It's likely the same travis-brokennes that also affected the master branch! Feel free to submit a PR now that you figured it out for master :wink:

sarahet:
> > But maybe we should find out why at some point ..
> 
> It's likely the same travis-brokennes that also affected the master branch! Feel free to submit a PR now that you figured it out for master

Although it's a different error message, but I will figure it out

PR #140, opened 2019-11-14
Add reduction from DNA5 to DNA4

sarahet:
Added nucleotide reduction from DNA5 to DNA4 with view convert_n. I decided to keep the view convert_n a deep view by definition since it does not make sense to apply it in any context to an outer range of ranges.

### DIFF
--- a/src/mkindex.hpp
+++ b/src/mkindex.hpp
@@ -55,7 +55,13 @@ void argConv2(LambdaIndexerOptions const & options);
 template <DbIndexType   c_indexType,
           AlphabetEnum  c_origAlph,
           AlphabetEnum  c_transAlph>
-void argConv3(LambdaIndexerOptions const & options);
+void argConv3a(LambdaIndexerOptions const & options);
+
+template <DbIndexType   c_indexType,
+          AlphabetEnum  c_origAlph,
+          AlphabetEnum  c_transAlph>
+void argConv3b(LambdaIndexerOptions const & options);
+
 
 template <DbIndexType           c_indexType,
           AlphabetEnum          c_origAlph,
@@ -129,7 +135,7 @@ void argConv1(LambdaIndexerOptions & options)
     {
 //         case AlphabetEnum::DNA4:            return argConv2<c_indexType, AlphabetEnum::DNA4>(options);
         case AlphabetEnum::DNA5:            return argConv2<c_indexType, AlphabetEnum::DNA5>(options);
-        case AlphabetEnum::AMINO_ACID:      return argConv3<c_indexType, AlphabetEnum::AMINO_ACID, AlphabetEnum::AMINO_ACID>(options);
+        case AlphabetEnum::AMINO_ACID:      return argConv3a<c_indexType, AlphabetEnum::AMINO_ACID, AlphabetEnum::AMINO_ACID>(options);
         default:                            throw 43;
     }
 }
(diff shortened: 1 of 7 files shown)


src/mkindex.hpp
```diff
@@ -43,6 +43,7 @@
 // Forwards
 // ==========================================================================
 
+
```
h-2:
```suggestion
```

src/mkindex.hpp
```diff
@@ -57,6 +58,12 @@ template <DbIndexType   c_indexType,
           AlphabetEnum  c_transAlph>
 void argConv3(LambdaIndexerOptions const & options);
 
+template <DbIndexType   c_indexType,
+          AlphabetEnum  c_origAlph,
+          AlphabetEnum  c_transAlph>
+void argConv4(LambdaIndexerOptions const & options);
```
h-2: Maybe rename these to `argConv3a` and `argConv3b` to indicate that they are called alternatively and not in succession?

src/search.hpp
```diff
@@ -51,11 +51,15 @@ template <DbIndexType   c_indexType,
           AlphabetEnum  c_origSbjAlph>
 void argConv2(LambdaOptions const & options);
 
+template <DbIndexType   c_indexType,
+          AlphabetEnum  c_origSbjAlph>
+void argConv3(LambdaOptions const & options);
```
h-2: Here, as well. If they are not called in succession, I would not number them increasingly. The whole thing is confusing enough as it is :wink:

src/shared_definitions.hpp
```diff
@@ -194,17 +195,26 @@ struct fake_index_file
...
+TRange & initHelper(TRange & input, TAdaptProt &&, TAdaptNucl &&)
 {
     return input;
 }
 
-template <typename TTargetAlph, typename TRange, typename TAdapt>
-auto initHelper(TRange & input, TAdapt && adapt)
+template <typename TTargetAlph, typename TRange, typename TAdaptProt, typename TAdaptNucl>
+    requires seqan3::NucleotideAlphabet<TTargetAlph> &&
+             !std::Same<seqan3::innermost_value_type_t<TRange>, TTargetAlph>
```
h-2:
```suggestion
             (!std::Same<seqan3::innermost_value_type_t<TRange>, TTargetAlph>)
```
As of GCC10 (and with Clang later) we need to put parens around requires expressions that are not "primary expressions" (i.e. everything that contains operators other than `&&` and `||`).

src/view_convert_n.hpp
```diff
@@ -0,0 +1,70 @@
+// ==========================================================================
+//                                  lambda
+// ==========================================================================
+// Copyright (c) 2013-2019, Hannes Hauswedell <h2 @ fsfe.org>
```
h-2: You can add your copyright whenever you make things :smile:

src/view_convert_n.hpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    template <std::ranges::Range urng_t>
+    constexpr friend auto operator|(urng_t && urange, convert_n_fn const & me)
+    {
+        return me(std::forward<urng_t>(urange));
+    }
+};
+
+inline constexpr auto convert_n = seqan3::view::deep{convert_n_fn{}};
```
h-2: Can you try the shorter version I posted in the chat?

h-2: I would put this item in namespace `views` for consistency with the others. And maybe make the name a little more expressive, e.g. `dna_n_to_random`? Names that end in `_n` have a certain different semantic associated with them, see 
https://en.cppreference.com/w/cpp/algorithm/copy_n
https://en.cppreference.com/w/cpp/algorithm/fill_n

And it might help to express in the name that it only works on dna or nucleotides.

h-2:
We need to remember to benchmark this threeway:

searching in DNA5
searching in DNA4 (with conversion and copy before)
searching in DNa4 with live-conversion via the view

src/view_convert_n.hpp
sarahet: Hm not sure this works, apparently the output of the shorter view version is not literal. We can evaluate this later maybe.

PR #142, opened 2019-11-22
[Fix] get correct ref names for sam/bam output

sarahet:
Former version had some problems:
1) Resize followed by append results in empty reference names in sam output as first n names are empty strings 
2) Iterating through substring appends single characters to stringset


sarahet:
Just another (unrelated) question: Is it due to SAM format specifics that the bit score decimal digits cannot be included in the tag?

PR #143, opened 2019-12-02
Adapt to new SeqAn3 version

sarahet:
Still fails during build and I can't figure out the problem :confused:

### DIFF
--- a/src/mkindex_algo.hpp
+++ b/src/mkindex_algo.hpp
@@ -23,9 +23,9 @@
 
 #include <seqan3/io/sequence_file/input.hpp>
 #include <seqan3/io/detail/misc_input.hpp>
-#include <seqan3/range/view/convert.hpp>
-#include <seqan3/range/view/translate.hpp>
-#include <seqan3/range/view/translate_join.hpp>
+#include <seqan3/range/views/convert.hpp>
+#include <seqan3/range/views/translate.hpp>
+#include <seqan3/range/views/translate_join.hpp>
 #include <seqan3/std/charconv>
 #include <seqan3/std/concepts>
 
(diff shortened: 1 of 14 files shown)


sarahet:
Now updated and works. Temporary solution as discussed: no bidirectional fm index.

src/mkindex_algo.hpp
```diff
@@ -691,38 +691,55 @@ auto parseAndStoreTaxTree(std::vector<bool>          & taxIdIsPresent,
...
+        //     case AlphabetEnum::AMINO_ACID: // no reduction
+        //     {
+        //         index = TIndex{seqs};
+        //     } break;
+        //
+        //     default:
+        //     {
+        //         throw std::runtime_error{"This reduced alphabet is not handled correctly."};
+        //     } break;
+        // }
```
h-2: I think all of this including the next line can just be replaced with 
```cpp
index = TIndex{seqs};
```
Or not?

sarahet: You mean including the reduced version?

h-2: Something like this:
```cpp
if constexpr (std::same_as<seqan3::innermost_value_type_t<TStringSet>, seqan3::TRedAlph>)
{
     index = TIndex{seqs};
}
else
{
    auto redSeqs = seqs | seqan3::views::deep{seqan3::views::convert<TRedAlph>};
    index = TIndex{redSeqs};
}
```

So there are no more alphabet specific things.

sarahet: Ahhh I see. Yes I think that should work, looks valid.

sarahet: Works and looks better now :slightly_smiling_face:

PR #145, opened 2019-12-16
[FIX] SIMD mode

h-2:
blocked by:

  *  ~~seqan/seqan#2410~~
  * ~~https://github.com/seqan/seqan3/pull/1449~~
  * ~~https://github.com/seqan/seqan/pull/2416~~

### DIFF
--- a/src/search_algo.hpp
+++ b/src/search_algo.hpp
@@ -357,7 +357,7 @@ loadQuery(GlobalDataHolder<c_indexType, c_origSbjAlph, c_transAlph, c_redAlph, c
     // TCDStringSet<std::vector<TOrigAlph>> origSeqs;
 
     // load
-    seqan3::sequence_file_input<QueryFileTraits<c_origQryAlph>, seqan3::fields<seqan3::field::ID, seqan3::field::SEQ>>
+    seqan3::sequence_file_input<QueryFileTraits<c_origQryAlph>, seqan3::fields<seqan3::field::id, seqan3::field::seq>>
         infile{options.queryFile};
 
 // TODO change to this once https://github.com/seqan/seqan3/issues/1135 is fixed
@@ -505,7 +505,7 @@ search(LocalDataHolder<TGlobalHolder> & lH)
     using TMatch = typename TGlobalHolder::TMatch;
 
     seqan3::configuration const cfg =
-        seqan3::search_cfg::max_error{seqan3::search_cfg::substitution{lH.options.maxSeedDist}};
+        seqan3::search_cfg::max_error{seqan3::search_cfg::substitution{static_cast<uint8_t>(lH.options.maxSeedDist)}};
 
     for (size_t i = lH.indexBeginQry; i < lH.indexEndQry; ++i)
     {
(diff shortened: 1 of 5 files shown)


src/search_algo.hpp
```diff
@@ -931,35 +927,34 @@ template <typename TLocalHolder>
...
-    using TBlastRecord  = BlastRecord<TBlastMatch,
-                                      decltype(lH.gH.qryIds[0]),
-                                      std::vector<std::string>,
-                                      decltype(lH.gH.indexFile.taxNames[0]),
-                                      uint32_t>;
+
+    using TBlastRecord  = seqan::BlastRecord<TBlastMatch,
+                                             seqan3::value_type_t<typename TGlobalHolder::TQryIds>,
+                                             std::vector<std::string>,
+                                             std::string, //seqan3::value_type_t<typename TGlobalHolder::TTaxNames>,
```
sarahet: Why did this change?

src/search_algo.hpp
```diff
@@ -931,35 +927,34 @@ template <typename TLocalHolder>
...
 //     using TPos          = typename TMatch::TPos;
     using TBlastPos     = uint32_t; //TODO why can't this be == TPos
-    using TBlastMatch   = BlastMatch<
+    using TBlastMatch   = seqan::BlastMatch<
                            typename TLocalHolder::TAlignRow0,
                            typename TLocalHolder::TAlignRow1,
                            TBlastPos,
-                           decltype(lH.gH.qryIds[0]),
-                           decltype(lH.gH.indexFile.subjIds[0]),
+                           seqan3::value_type_t<typename TGlobalHolder::TQryIds>,// TODO: reference_t?
```
sarahet: Why the ```todo```? Why doesn't it work now?

src/seqan2seqan3.hpp
```diff
@@ -246,7 +259,41 @@ inline auto score(Score<TValue, TSpec> const & scheme, alph_t const a1, alph_t c
...
+    {
+        return seqan3::to_rank(c);
+    }
+};
 
+template <typename input_t>
+struct seqan2_to_rank_outer
+{
+    using argument_type = typename GetValue<input_t const>::Type;
+    using result_type = seqan::ModifiedString<std::remove_reference_t<argument_type> const, seqan::ModView<seqan2_to_rank_inner>>;
```
sarahet: Introduce line break

h-2:
> I guess most of the changes are still just adapting to our new SeqAn3/SeqAn2 version of lambda. 

Yep!

> However, for the new objects you introduced in order to fix the SIMD version, could you add a bit of description/documentation in the code? 

Which new objects?

src/search_algo.hpp
h-2: I have changed this now to use views everywhere. It required a few more changes but avoid useless copies.

src/seqan2seqan3.hpp
h-2: done

sarahet:
> Which new objects?

I meant everything that is new in the seqan2seqan3 header file.

h-2:
> I meant everything that is new in the seqan2seqan3 header file.

Ok, I can add some comments, but it's all just glue for bridging between SeqAn2 and SeqAn3. I can tell you why this specific stuff is necessary but I also just add things based on demand so the list of types and weird workaround will likely grow.

The real solution is to update fully to SeqAn3 once that has SIMD.

sarahet:
> The real solution is to update fully to SeqAn3 once that has SIMD.

I agree, but if this takes longer I might need some quick info on what the workarounds do :slightly_smiling_face:

src/search_algo.hpp
```diff
@@ -869,27 +867,25 @@ _performAlignment(TDepSetH & depSetH,
...
+    using TAlignConfig  = seqan::AlignConfig2<seqan::LocalAlignment_<>,
+                                       seqan::DPBandConfig<seqan::BandOff>,
+                                       seqan::FreeEndGaps_<seqan::True, seqan::True, seqan::True, seqan::True>,
                                        std::conditional_t<withTrace,
-                                                          TracebackOn<TracebackConfig_<CompleteTrace, GapsLeft> >,
-                                                          TracebackOff> >;
-    using TSimdAlign    = typename SimdVector<int16_t>::Type;
-    using TSimdScore    = Score<TSimdAlign, ScoreSimdWrapper<typename TGlobalHolder::TScoreScheme> >;
-    using TSize         = typename Size<typename TLocalHolder::TAlignRow0>::Type;
+                                                          seqan::TracebackOn<seqan::TracebackConfig_<seqan::CompleteTrace, seqan::GapsLeft> >,
```
sarahet: line break

PR #148, opened 2020-01-07
Various fixes

h-2:
~~blocked by seqan/seqan#2418~~


sarahet:
Is this now still not comparable to lambda2 SIMD?

h-2:
Just FYI, they broke some CMAKE stuff in SeqAn master, that's why our CI is failing.

src/search_datastructures.hpp
```diff
@@ -408,11 +408,6 @@ class GlobalDataHolder
     using TQryIds       = TIds;
     using TSubjIds      = TIds;
 
-    /* indeces and their type */
-    using TDbIndex      = std::conditional_t<c_dbIndexType == DbIndexType::BI_FM_INDEX,
```
sarahet: Why don't we need this anymore? I thought we will include the bidirectional index again at some point.

src/search_algo.hpp
```diff
@@ -1166,9 +1165,9 @@ iterateMatchesFullSerial(TLocalHolder & lH)
         bm._n_qId = it->qryId / seqan::qNumFrames(TGlobalHolder::blastProgram);
         bm._n_sId = it->subjId / seqan::sNumFrames(TGlobalHolder::blastProgram);
 
-        bm.sLength = seqan::sIsTranslated(TGlobalHolder::blastProgram)
-                        ? lH.gH.indexFile.seqs[it->subjId].size()
-                        : lH.gH.transSbjSeqs[it->subjId].size();
+        bm.qLength = //std::ranges::size(lH.gH.transQrySeqs[it->qryId]);
```
sarahet: What about the comment? Is this needed in the future?

src/search_datastructures.hpp
h-2: This alias is just a leftover of lambda2, it's independent of everything else (the index type is now part of the index file that we use as-is).

src/search_algo.hpp
h-2: We should discuss this again. Like I said, I think basing the variable on the actual translated seq length would be better for the e-value calculation but for now the focus was on reproducing lambda2's results.

PR #150, opened 2020-01-22
[Fix] fix alphabet detection and subject lengths for tblastn and tblastx

sarahet:


src/shared_misc.hpp
```diff
@@ -118,21 +118,29 @@ AlphabetEnum detectSeqFileAlphabet(std::string const & path)
 
     auto & seq = std::get<0>(*f.begin());
 
-    if (std::ranges::equal(seq | seqan3::views::char_to<seqan3::dna5> | seqan3::views::to_char,
-                           seq | seqan3::views::to_upper))
+    auto u_to_t_view = std::views::transform([] (auto const in)
+    {
+        if (in == 'U')
```
h-2: Hm, lower-case U is sill missing. I think a better solution to the whole thing is adding this small function

```cpp
template <typename TAlph, typename TRange>
constexpr bool all_valid(TRange && r)
{
    for (auto const c : r)
        if (!seqan3::char_is_valid_for<TAlph>(c))
            return false;
    return true;
}
```

And then just check `if (all_valid<seqan3::dna5>(seq))`  and likewise for the other alphabets.

h-2: (U is explicitly defined as a valid char for the DNAs in SeqAn3)

PR #151, opened 2020-01-23
[FEATURE] re-activate bi-fm-index

h-2:
This PR re-activates the BI-FM-Index. It contains a workaround for a build-failure in GCC7 (https://github.com/seqan/seqan3/pull/1519).

But at least on my systems, this still ICEs on GCC8 and GCC9. Can you verify this, @sarahet ?

I have reported this here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92778

Some preliminary tests with bifm (using wavelet trees) are promising (20% raw speed-up, possibly more with different strategies).


sarahet:
I cannot reproduce this, the PR builds on my system with gcc8 and gcc9.

PR #152, opened 2020-01-27
[FEATURE] add li10 alphabet

h-2:
First results are looking promising, but let's wait for bigger files.


src/search_options.hpp
```diff
@@ -162,7 +162,7 @@ void parseCommandLine(LambdaOptions & options, int argc, char const ** argv)
 
     // TODO Better solution for file extensions
     parser.add_option(options.queryFile, 'q', "query", "Query sequences.", seqan3::option_spec::REQUIRED,
-        seqan3::input_file_validator{{"fa", "fasta", "fq", "fastq", "gz"}});
+        seqan3::input_file_validator<seqan3::sequence_file_input<>>{});
```
h-2: This change is unrelated and just automates getting the extensions.

h-2:
(It does increase compile-time even further :expressionless: )

PR #153, opened 2020-01-28
[FEATURE] make adaptive seeding available again

h-2:
Making the internal search ourselves has quite positive effect on performance. 

Current version of adaptive seeding is very conservative and only brings 0-10% speed improvement (but it also doesn't make anything worse).

Splitting the translation units reduces compile time drastically again. Attention: if review the commits individually the diff is correct and shows files being renamed, but if you look at the total diff, it looks like huge copy'n'paste.

### DIFF
--- a/src/search_algo.hpp
+++ b/src/search_algo.hpp
@@ -497,6 +497,24 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
     return false;
 }
 
+/*** THIS ACCESSES seqan3::detail:: IT SHOULD BE REPLACED WITH OFFICIAL INTERFACE ONCE OPTIMISED ***/
+template <typename TGlobalHolder, typename TSeed>
+inline void
+search_impl(LocalDataHolder<TGlobalHolder> & lH, TSeed && seed)
+{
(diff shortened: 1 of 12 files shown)


src/search_algo.hpp
```diff
@@ -497,15 +497,45 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
     return false;
 }
 
+/*** THIS ACCESSES seqan3::detail:: IT SHOULD BE REPLACED WITH OFFICIAL INTERFACE ONCE OPTIMISED ***/
```
sarahet: Do I understand it correctly that we use the detail functions to enable adaptive seeding or is there a second problem where the general search is too slow (and why)?

h-2: This is entirely unrelated, adaptive seeding happens only via
https://github.com/seqan/lambda/pull/153/commits/c241387b97808d011c531d3ef70c6313144c1972#diff-ee6754a210a5bb744bff645ba899ae8dR570

The function you marked circumenvents various unecessary memory allocations in `seqan3::search`. See https://github.com/seqan/seqan3/pull/1528 for more details.

sarahet: Ah thanks for clarifying this.

PR #155, opened 2020-01-29
Stream query

h-2:
blocked by #154 

(only the last commit is relevant is specific to this PR)

Includes progress bar again, although I haven't checked how fast counting is; likely room for optimisation.

### DIFF
--- a/src/search_algo.hpp
+++ b/src/search_algo.hpp
@@ -329,122 +329,34 @@ loadDbIndexFromDisk(GlobalDataHolder<c_indexType, c_origSbjAlph, c_transAlph, c_
 // Function loadQuery()
 // --------------------------------------------------------------------------
 
-template <AlphabetEnum c_origQryAlph>
-struct QueryFileTraits : std::conditional_t<c_origQryAlph == AlphabetEnum::DNA5,
-                                            seqan3::sequence_file_input_default_traits_dna,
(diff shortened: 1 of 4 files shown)


sarahet:
All builds fail ..

src/search_algo.hpp
```diff
@@ -1215,10 +1125,10 @@ iterateMatchesFullSerial(TLocalHolder & lH)
 
     auto const trueQryId = lH.matches[0].qryId / qNumFrames(TGlobalHolder::blastProgram);
 
-    TBlastRecord record(lH.gH.qryIds[trueQryId]);
+    TBlastRecord record(lH.qryIds[trueQryId]);
     record.qLength = seqan::length(lH.gH.qrySeqs[trueQryId]);
```
sarahet: This needs to be ```record.qLength = seqan::length(lH.qrySeqs[trueQryId]);```

h-2: Yes, that's artifact from merging the branches.

sarahet:
Since we talked about divergence of branches, should be maybe address this here as well and account for the recent changes in the master branch?

h-2:
> Since we talked about divergence of branches, should be maybe address this here as well and account for the recent changes in the master branch?

No, we should merge master into lambda3 at some point and resolve the conflicts. But it doesn't have to happen in the next days.

PR #156, opened 2020-01-30
[Fix] Get correct number of header entries for subject sequences

sarahet:
For translated databases, the number of original sequence lengths was used to determine the header entries. The last entry of the original sequence lengths is however the sum, so one empty header line was accidentally created in this case.


sarahet:
> LGTM, but does this intentionally only target the master branch (lambda2)?

Yes, only a problem in lambda2 as we directly access the sequences and not a length vector in lambda3. In lambda3 is a different bug in the bam header, though, that will be fixed by #157.

h-2:
Note that we have increasing divergence between the branches. At some point, this needs to be synchronised again.

PR #157, opened 2020-01-30
[Fix] correct CIGAR string in SAM format as well as correct header in…

sarahet:
… bam format

Already known problem with macth/record qLength entries.
Wrong number of sequences used for the BAM header


src/search_output.hpp
```diff
@@ -353,9 +352,9 @@ myWriteHeader(TGH & globalHolder, TLambdaOptions const & options)
         auto & subjIds          = seqan::contigNames(context);
 
         // compute seqan::lengths ultra-fast
-        seqan::resize(subjSeqLengths, globalHolder.redSbjSeqs.size());
+        seqan::resize(subjSeqLengths, globalHolder.indexFile.seqs.size());
         SEQAN_OMP_PRAGMA(parallel for simd)
-        for (size_t i = 0; i < globalHolder.redSbjSeqs.size(); ++i)
+        for (size_t i = 0; i < globalHolder.indexFile.seqs.size(); ++i)
```
h-2: wow, good catch, this looks really weird!

PR #158, opened 2020-02-03
fix block sizes

h-2:


src/search_datastructures.hpp
```diff
@@ -430,6 +430,12 @@ class GlobalDataHolder
     size_t              queryTotal = 0;
     std::atomic<size_t> queryCount = 0;
     size_t              records_per_batch = 100;
+
+    GlobalDataHolder() = default;
```
sarahet: Is there a particular reason why we did not have them before?

h-2: The compiler generates defaults. I just thought it's good to explicitly add them and delete the ones that shouldn't be used anyway.

PR #160, opened 2020-02-17
[Feature] Add bisulfite reduction

sarahet:
This is not completely ready yet as I am not sure about the final measurements. Also new index test files are missing so please do not merge yet. 

@h-2 Could you anyways have a first look and see whether this looks generally fine? First tests worked but not sufficient yet.


src/bisulfite_scoring.hpp
```diff
@@ -0,0 +1,133 @@
...
+        if ((i_ms  < std::numeric_limits<score_type>::lowest() || i_ms  > std::numeric_limits<score_type>::max()) ||
+            (i_mms < std::numeric_limits<score_type>::lowest() || i_mms > std::numeric_limits<score_type>::max()))
+        {
+            throw std::invalid_argument{"You passed a score value to set_bisulfite_scheme that is out of range of the "
+                                        "scoring scheme's underlying type. Define your scoring scheme with a larger "
+                                        "template parameter or down-cast you score value beforehand to prevent "
+                                        "this exception."};
+        }
+
+        // Assume query is horizontal sequence
```
h-2: I would maybe override the score function from `scoring_scheme_base` to ensure that exactly one parameter is of type `dna3bs` and the other isn't? Seems like it is easy to get wrong otherwise...

src/bisulfite_scoring.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+#include <range/v3/algorithm/copy.hpp>
+
+#include <seqan3/alignment/scoring/scoring_scheme_base.hpp>
+#include <seqan3/alphabet/nucleotide/dna5.hpp>
+#include <seqan3/range/shortcuts.hpp>
+#include <seqan3/std/algorithm>
+
+namespace seqan3
+{
```
h-2: This should just be in Lambda's namespace. [SeqAn3 actually states that you should never open namespace seqan3 in your app :wink: ]

src/search_algo.hpp
```diff
@@ -196,9 +196,28 @@ prepareScoring(GlobalDataHolder<c_dbIndexType, c_origSbjAlph, c_transAlph, c_red
...
+
+            // Seqan3
+            globalHolder.scoringScheme.set_bisulfite_scheme(seqan3::match_score{options.match},
+                                                            seqan3::mismatch_score{options.misMatch});
+        }
+        else
+        {
+            // Seqan2
+            seqan::setScoreMatch(globalHolder.scoringSchemeAlign, options.match);
+            seqan::setScoreMismatch(globalHolder.scoringSchemeAlign, options.misMatch);
```
h-2: replace with `globalHolder.scoringSchemeAlign = context(globalHolder.outfileBlastTab)` ?

src/search_algo.hpp
```diff
@@ -196,9 +196,28 @@ prepareScoring(GlobalDataHolder<c_dbIndexType, c_origSbjAlph, c_transAlph, c_red
...
+
+            // Seqan3
+            globalHolder.scoringScheme.set_simple_scheme(seqan3::match_score{options.match},
+                                                         seqan3::mismatch_score{options.misMatch});
```
h-2: same lines before if-body. Are both required? Where is `globalHolder.scoringScheme` currently used at all?

src/search_datastructures.hpp
```diff
@@ -396,14 +398,24 @@ class GlobalDataHolder
...
     /* output file */
     using TScoreScheme3  = std::conditional_t<seqan3::nucleotide_alphabet<TRedAlph>,
-                                              seqan3::nucleotide_scoring_scheme<>,
+                                              std::conditional_t<c_redAlph == AlphabetEnum::DNA3BS,
+                                                                 seqan3::bisulfite_scoring_scheme<>,
+                                                                 seqan3::nucleotide_scoring_scheme<>>,
                                               seqan3::aminoacid_scoring_scheme<>>;
 
-    using TScoreScheme  =
+    using TScoreSchemeAlign  =
```
h-2: Maybe add comments on what these are for?

src/seqan2seqan3.hpp
```diff
@@ -268,6 +271,14 @@ inline auto score(Score<TValue, ScoreMatrix<TSequenceValue, TSpec>> const & sche
     return score(scheme, AminoAcid{seqan3::to_char(a1)}, AminoAcid{seqan3::to_char(a2)});
 }
 
+template <typename TValue, seqan3::nucleotide_alphabet alph_t>
+inline auto score(Score<TValue, ScoreMatrix<Dna5, BisulfiteMatrix>> const & scheme,
+                  alph_t const a1,
+                  alph_t const a2)
+{
+    return score(scheme, Dna5{seqan3::to_char(a1)}, Dna5{seqan3::to_char(a2)});
```
h-2: Are both intentionally converted to Dna5?

src/view_dna_n_to_random.hpp
```diff
@@ -49,11 +51,11 @@ struct dna_n_to_random_fn
         {
             if (seqan3::to_char(in) == 'N')
             {
-                return seqan3::assign_rank_to(rng() % 4, seqan3::dna4{});
+                return seqan3::assign_rank_to(rng() % TAlph::alphabet_size, TAlph{});
```
h-2:
```suggestion
                return seqan3::assign_rank_to(rng() % seqan3::alphabet_size<TAlph>, TAlph{});
```

Please use global traits in generic contexts.

src/bisulfite_scoring.hpp
sarahet: Well, I must have missed this :stuck_out_tongue_winking_eye:

src/search_algo.hpp
sarahet: I am unsure this would work .. Doesn't it require the bisulfite scheme to work with all the blast functions that allow only simple scheme?

src/search_algo.hpp
sarahet: For the scoring in the seed looks promising function? Or not?

src/seqan2seqan3.hpp
sarahet: I made the bisulfite scheme for Dna5 sequences .. I thought this needs to be done then?

src/search_algo.hpp
h-2: :thinking: 

This if-case is not about BS at all

src/search_algo.hpp
h-2: > For the scoring in the seed looks promising function? Or not?

Ah, yes!

>> same lines before if-body. Are both required?

This is still open.

src/seqan2seqan3.hpp
h-2: Ah, true. Scoring happens on unreduced and that is always Dna5.

src/search_algo.hpp
sarahet: Oh right, I checked the wrong line :see_no_evil:

src/search_algo.hpp
sarahet: Yes, only the one in the if clause is required, must be an accidental leftover from copy/pasting

src/search_algo.hpp
sarahet: Anyways, this does not work as one is a Blast scoring scheme and the other is not ..

sarahet:
I addressed all your comments. However, I still have problems with the SIMD mode but it seems not to be an issue with the SIMD implementation itself. I made a small working example [here](https://gist.github.com/sarahet/97ca17c402d4a318beeac098319aa1de) which gives me the expected result of accepted T to C mismatches. However, the SIMD mode in lambda does not output these alignments instead shortens it. The example sequence from the gist are directly taken from an example I ran through lambda and the sequences that go into the alignment there. I saw that lambda SIMD mode does just not compute the alignment it should (but the gist example does). Do you have any idea what could be different between the two?

sarahet:
Took forever but bug found and fixed. Do you think there would be any problem with the current fix in seqan2seqan3 for nucleotide mode without bilsufite scoring? I did not see a way of constraining this easily to only bisulfite mode since both work on dna5 in the alignment step and so far there was no problem when testing the nucleotide mode. Are there anyways any other alphabets than dna5 and aa27 that can go into this function? Otherwise we could just omit the if-statement completely.

src/search_datastructures.hpp
```diff
@@ -423,6 +438,7 @@ class GlobalDataHolder
     TBamFile            outfileBam;
 
     TScoreScheme3       scoringScheme;
```
h-2:
```suggestion
    TScoreScheme3       scoringSchemePreScoring;
```
Maybe denote clearly what this is used for?

src/seqan2seqan3.hpp
```diff
@@ -340,6 +351,13 @@ inline constexpr std::array<uint8_t, 27> seqan3rank_to_seqan2rank<seqan3::aa27>
 { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 25, 23, 24, 26 };
 //A  B  C  D  E  F  G  H  I  J  K   L   M   N   O   P   Q   R   S   T   U   V   W   Y   Z   X   *       <- seqan2
 //                                                                                  ↑   ↑   ↑
+// aa27 are not:
```
h-2:
```suggestion
// dna5 is scored via matrix in bisulphite-mode
```

src/seqan2seqan3.hpp
```diff
@@ -362,7 +380,7 @@ struct seqan2_to_rank_inner
     template <typename t>
     constexpr uint8_t operator()(t const c) const noexcept
     {
-        if constexpr (std::is_same_v<t, seqan3::aa27> /* or dna3bs*/)
+        if constexpr (std::is_same_v<t, seqan3::aa27> || std::is_same_v<t, seqan3::dna5>)
```
h-2:
```suggestion
        if constexpr (std::is_same_v<t, seqan3::aa27> || std::is_same_v<t, seqan3::dna5>) // dna5 used in bisulphite-mode
```

h-2: This is unnecessary for regular dna-alignment, but all "fixes" would be quite complicated, so let's leave it like this.

src/view_dna_n_to_random.hpp
```diff
@@ -29,6 +30,7 @@
 #include <seqan3/range/views/deep.hpp>
 #include <seqan3/std/ranges>
 
+template <typename TAlph>
```
h-2:
```suggestion
template <seqan3::nucleotide_alphabet TAlph>
```

Maybe restrict to only nucleotides? Otherwise, 'N' has no clear meaning.

src/view_dna_n_to_random.hpp
```diff
@@ -68,6 +70,7 @@ struct dna_n_to_random_fn
 namespace seqan3::views
 {
 
-inline constexpr auto dna_n_to_random = seqan3::views::deep{dna_n_to_random_fn{}};
+template<typename TAlph>
```
h-2:
```suggestion
template<typename TAlph = seqan3::dna4>
```

So dna4 is still the default and doesn't have to specified.

h-2:
P.S: please do not mix updates of submodules with other changes in a single commit !

sarahet:
> P.S: please do not mix updates of submodules with other changes in a single commit !

Oh sorry. Can I somehow reverse this?

src/view_dna_n_to_random.hpp
```diff
@@ -68,6 +70,7 @@ struct dna_n_to_random_fn
 namespace seqan3::views
 {
 
-inline constexpr auto dna_n_to_random = seqan3::views::deep{dna_n_to_random_fn{}};
+template<typename TAlph = seqan3::dna4>
```
h-2, line 73:
```suggestion
template <typename TAlph = seqan3::dna4>
```

Just this minor remark, feel free to merge after fixing it!

PR #166, opened 2021-02-12
[Feature] Add second bisulfite index

sarahet:
First PR in order to add the second bisulfite index. Not finished yet (only serial mode currently working).

### DIFF
--- a/src/search.cpp
+++ b/src/search.cpp
@@ -16,7 +16,7 @@
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 //
 // ==========================================================================
-// lambda.cpp: Main File for Lambda
+// search.cpp: Main File for the search
 // ==========================================================================
 
 #include "seqan2seqan3.hpp" // must come first
@@ -369,16 +369,6 @@ void realMain(LambdaOptions     const & options)
                 localHolder.qryIds.push_back(std::move(id));
                 localHolder.qrySeqs.push_back(std::move(seq));
 
-                //TODO this is not ideal:
-                if constexpr (c_transAlph == AlphabetEnum::DNA5) // BLASTN
-                {
-                    auto revComp = localHolder.qrySeqs.back()
-                                 | std::views::reverse
-                                 | seqan3::views::complement
-                                 | seqan3::views::to<std::ranges::range_value_t<decltype(localHolder.qrySeqs)>>;
-                    localHolder.qrySeqs.push_back(std::move(revComp));
-                }
-
                 if (++localHolder.queryCount == globalHolder.records_per_batch)
                     break;
             }
@@ -398,7 +388,6 @@ void realMain(LambdaOptions     const & options)
         #ifdef LAMBDA_MICRO_STATS
             localHolder.stats.timeSearch += sysTime() - buf;
         #endif
-
             // extend
             if (localHolder.matches.size() > 0)
                 iterateMatches(localHolder);
--- a/src/search_algo.hpp
+++ b/src/search_algo.hpp
@@ -16,7 +16,7 @@
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 //
 // ==========================================================================
-// lambda.hpp: contains the main progam pipeline
+// search_algo.hpp: functions to set up, perform and evaluate the search
 // ==========================================================================
 
 #pragma once
@@ -36,6 +36,7 @@
 
 #include <seqan/align_extend.h>
 
+#include <seqan3/core/debug_stream.hpp>
 #include <seqan3/range/views/complement.hpp>
 #include <seqan3/range/views/translate_join.hpp>
 #include <seqan3/io/sequence_file/input.hpp>
@@ -43,6 +44,9 @@
 #include <seqan3/alphabet/aminoacid/aa27.hpp>
 
 #include "bisulfite_scoring.hpp"
+#include "evaluate_bisulfite_alignment.hpp"
+#include "view_dna_n_to_random.hpp"
+#include "view_reduce_to_bisulfite.hpp"
 
 // ============================================================================
 // Forwards
@@ -199,19 +203,25 @@ prepareScoring(GlobalDataHolder<c_dbIndexType, c_origSbjAlph, c_transAlph, c_red
         if constexpr (c_redAlph == AlphabetEnum::DNA3BS)
         {
             // Seqan2
-            seqan::setScoreBisulfiteMatrix(globalHolder.scoringSchemeAlign, options.match, options.misMatch);
+            seqan::setScoreBisulfiteMatrix(globalHolder.scoringSchemeAlign, options.match, options.misMatch, bsDirection::fwd);
+            seqan::setScoreBisulfiteMatrix(globalHolder.scoringSchemeAlignBSRev, options.match, options.misMatch, bsDirection::rev);
 
             // Seqan3
             globalHolder.scoringSchemePreScoring.set_bisulfite_scheme(seqan3::match_score{options.match},
-                                                            seqan3::mismatch_score{options.misMatch});
+                                                         seqan3::mismatch_score{options.misMatch});
+            globalHolder.scoringSchemePreScoringBSRev.set_bisulfite_scheme(seqan3::match_score{options.match},
+                                                         seqan3::mismatch_score{options.misMatch}, bsDirection::rev);
         }
         else
         {
             // Seqan2
             globalHolder.scoringSchemeAlign = seqan::seqanScheme(context(globalHolder.outfileBlastTab).scoringScheme);
+            globalHolder.scoringSchemeAlignBSRev = seqan::seqanScheme(context(globalHolder.outfileBlastTab).scoringScheme);
 
             // Seqan3
             globalHolder.scoringSchemePreScoring.set_simple_scheme(seqan3::match_score{options.match},
+                                                         seqan3::mismatch_score{options.misMatch});
+            globalHolder.scoringSchemePreScoringBSRev.set_simple_scheme(seqan3::match_score{options.match},
                                                          seqan3::mismatch_score{options.misMatch});
         }
     }
@@ -243,8 +253,11 @@ prepareScoring(GlobalDataHolder<c_dbIndexType, c_origSbjAlph, c_transAlph, c_red
                                   seqan2_matrix_id);
         seqan::setScoreMatrixById(globalHolder.scoringSchemeAlign,
                                   seqan2_matrix_id);
+        seqan::setScoreMatrixById(globalHolder.scoringSchemeAlignBSRev,
+                                  seqan2_matrix_id);
         // Seqan3
         globalHolder.scoringSchemePreScoring.set_similarity_matrix(seqan3_matrix_id);
+        globalHolder.scoringSchemePreScoringBSRev.set_similarity_matrix(seqan3_matrix_id);
     }
 
     // seqan2
@@ -253,6 +266,9 @@ prepareScoring(GlobalDataHolder<c_dbIndexType, c_origSbjAlph, c_transAlph, c_red
 
     seqan::setScoreGapOpen(globalHolder.scoringSchemeAlign, options.gapOpen + options.gapExtend);
     seqan::setScoreGapExtend(globalHolder.scoringSchemeAlign, options.gapExtend);
+
+    seqan::setScoreGapOpen(globalHolder.scoringSchemeAlignBSRev, options.gapOpen + options.gapExtend);
+    seqan::setScoreGapExtend(globalHolder.scoringSchemeAlignBSRev, options.gapExtend);
 
     if (!seqan::isValid(seqan::context(globalHolder.outfileBlastTab).scoringScheme))
         throw std::runtime_error{"Could not compute Karlin-Altschul-Values for Scoring Scheme.\n"};
@@ -294,10 +310,12 @@ loadDbIndexFromDisk(GlobalDataHolder<c_indexType, c_origSbjAlph, c_transAlph, c_
         throw 88;
     }
 
-    if constexpr (c_origSbjAlph == c_transAlph)
-        globalHolder.transSbjSeqs = globalHolder.indexFile.seqs | seqan3::views::type_reduce;       // no-op view
-    else
+    if constexpr (c_redAlph == AlphabetEnum::DNA3BS)
+        globalHolder.transSbjSeqs = globalHolder.indexFile.seqs | views::duplicate;
+    else if constexpr (c_origSbjAlph != c_transAlph)
         globalHolder.transSbjSeqs = globalHolder.indexFile.seqs | seqan3::views::translate_join;
+    else
+        globalHolder.transSbjSeqs = globalHolder.indexFile.seqs | seqan3::views::type_reduce;       // no-op view
 
     if constexpr (c_transAlph == c_redAlph)
     {
@@ -307,8 +325,17 @@ loadDbIndexFromDisk(GlobalDataHolder<c_indexType, c_origSbjAlph, c_transAlph, c_
     {
         if constexpr (c_transAlph == AlphabetEnum::DNA5) // BLASTN mode
         {
-            globalHolder.redSbjSeqs = globalHolder.transSbjSeqs
-                                    | seqan3::views::dna_n_to_random<_alphabetEnumToType<c_redAlph>>;
+            if constexpr (c_redAlph == AlphabetEnum::DNA3BS) // BLASTN mode
+            {
+                globalHolder.redSbjSeqs = globalHolder.transSbjSeqs
+                                        | views::dna_n_to_random<seqan3::dna4>
+                                        | views::reduce_to_bisulfite;
+            }
+            else
+            {
+                globalHolder.redSbjSeqs = globalHolder.transSbjSeqs
+                                        | views::dna_n_to_random<_alphabetEnumToType<c_redAlph>>;
+            }
         }
         else
         {
@@ -382,7 +409,7 @@ countQuery(GlobalDataHolder<c_indexType, c_origSbjAlph, c_transAlph, c_redAlph, 
 /// THREAD LOCAL STUFF
 
 // --------------------------------------------------------------------------
-// Function generateSeeds()
+// Function seedLooksPromising()
 // --------------------------------------------------------------------------
 
 // perform a fast local alignment score calculation on the seed and see if we
@@ -428,9 +455,13 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
     int const thresh  = lH.options.preScoringThresh * effectiveLength;
 
     // score the diagonal
+    auto & currentScoringScheme = TGlobalHolder::c_redAlph == AlphabetEnum::DNA3BS && m.subjId % 2 ?
+                                  lH.gH.scoringSchemePreScoringBSRev :  lH.gH.scoringSchemePreScoring;
+
     for (uint64_t i = 0; i < effectiveLength; ++i)
     {
-        s += lH.gH.scoringSchemePreScoring.score(qSeq[i], sSeq[i]);
+        s += currentScoringScheme.score(qSeq[i], sSeq[i]);
+
         if (s < 0)
             s = 0;
         else if (s > maxScore)
@@ -439,7 +470,6 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
         if (maxScore >= thresh)
             return true;
     }
-
     return false;
 }
 
@@ -507,7 +537,6 @@ search(LocalDataHolder<TGlobalHolder> & lH)
             // results are in cursor_buffer
             lH.cursor_buffer.clear();
             search_impl(lH, lH.redQrySeqs[i] | seqan3::views::slice(seedBegin, seedBegin + lH.options.seedLength));
-
             for (auto & cursor : lH.cursor_buffer)
             {
                 size_t seedLength = lH.options.seedLength;
@@ -966,13 +995,13 @@ iterateMatchesFullSimd(TLocalHolder & lH)
     for (auto it = lH.matches.begin(), itEnd = lH.matches.end(); it != itEnd; ++it)
     {
         // create blastmatch in list without copy or move
-        blastMatches.emplace_back(lH.qryIds [it->qryId / seqan::qNumFrames(TGlobalHolder::blastProgram)],
-                                  lH.gH.indexFile.ids[it->subjId / seqan::sNumFrames(TGlobalHolder::blastProgram)]);
+        blastMatches.emplace_back(lH.qryIds [it->qryId / TGlobalHolder::qryNumFrames],
+                                  lH.gH.indexFile.ids[it->subjId / TGlobalHolder::sbjNumFrames]);
 
         auto & bm = seqan::back(blastMatches);
 
-        bm._n_qId = it->qryId / seqan::qNumFrames(TGlobalHolder::blastProgram);
-        bm._n_sId = it->subjId / seqan::sNumFrames(TGlobalHolder::blastProgram);
+        bm._n_qId = it->qryId / TGlobalHolder::qryNumFrames;
+        bm._n_sId = it->subjId / TGlobalHolder::sbjNumFrames;
 
         bm.qLength = //std::ranges::size(lH.transQrySeqs[it->qryId]);
                     std::ranges::size(lH.qrySeqs[bm._n_qId]);
@@ -1143,12 +1172,12 @@ iterateMatchesFullSerial(TLocalHolder & lH)
                                              std::string_view,
                                              uint32_t>;
 
-    auto const trueQryId = lH.matches[0].qryId / qNumFrames(TGlobalHolder::blastProgram);
+    auto const trueQryId = lH.matches[0].qryId / TGlobalHolder::qryNumFrames;
 
     TBlastRecord record(lH.qryIds[trueQryId]);
     record.qLength = seqan::length(lH.qrySeqs[trueQryId]);
 
-    size_t band = _bandSize(seqan::length(lH.transQrySeqs[lH.matches[0].qryId]), lH);
+    size_t band = _bandSize(lH.transQrySeqs[lH.matches[0].qryId].size(), lH);
 
 #ifdef LAMBDA_MICRO_STATS
     double start = sysTime();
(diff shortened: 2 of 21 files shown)


src/search_algo.hpp
```diff
@@ -443,10 +481,87 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
...
+                                           0};                                           // deletions
+
+    auto delegate = [&lH] (auto const & it)
+    {
+        lH.cursor_buffer_fwd.push_back(it);
+    };
+
+    seqan3::detail::search_algo<false>(lH.gH.indexFileFwd.index, seed, max_error, delegate);
+}
+
```
h-2, line 560: I would really like to prevent this duplication of logic here. If the same type is used for the scoring scheme (just two different objects), one could simply use one `seedLooksPromising()` function and add this at the beginning:
```cpp
auto const & scoringScheme = runtime_variable ? lH.gH.scoringSchemePreScoringRev :  lH.gH.scoringSchemePreScoringFwd;
```
And then use that later on.

h-2: Especially since this function works with transSeqs and not redSeqs

src/search_algo.hpp
```diff
@@ -542,7 +657,99 @@ search(LocalDataHolder<TGlobalHolder> & lH)
...
+                        lH.matches.push_back(m);
+                }
+            }
+        }
+    }
+}
+
+template <typename TGlobalHolder>
+inline void
+search_rev(LocalDataHolder<TGlobalHolder> & lH)
```
h-2, line 672: Please implement this via an extra template argument to avoid copy'n'pasting the entire function!

src/search_algo.hpp
```diff
@@ -889,7 +1096,7 @@ _performAlignment(TDepSetH & depSetH,
     unsigned constexpr sizeBatch = seqan::LENGTH<TSimdAlign>::VALUE;
     unsigned const      fullSize = sizeBatch * ((seqan::length(blastMatches) + sizeBatch - 1) / sizeBatch);
 
-    TSimdScore simdScoringScheme(lH.gH.scoringSchemeAlign);
+    TSimdScore simdScoringScheme(lH.gH.scoringSchemeAlignFwd);
```
h-2, line 1099: Why are there two versions of the scoringSchemeAlign? Isn't that only used in Alignment (which happens on unreduced seqs)?

src/search_datastructures.hpp
```diff
@@ -339,10 +341,11 @@ class GlobalDataHolder
     static constexpr AlphabetEnum  c_redAlph       = c_redAlph_;
     static constexpr AlphabetEnum  c_origQryAlph   = c_origQryAlph_;
 
-    using TOrigQryAlph   = _alphabetEnumToType<c_origQryAlph>;
-    using TOrigSbjAlph   = _alphabetEnumToType<c_origSbjAlph>;
-    using TTransAlph     = _alphabetEnumToType<c_transAlph>;
-    using TRedAlph       = _alphabetEnumToType<c_redAlph>;
+    using TOrigQryAlph   = _alphabetEnumToType<c_origQryAlph, true>;
+    using TOrigSbjAlph   = _alphabetEnumToType<c_origSbjAlph, true>;
+    using TTransAlph     = _alphabetEnumToType<c_transAlph, true>;
```
h-2, line 346: Why do these get an extra parameter, it is always just the reduced alphabet that can change, or not?

src/search_datastructures.hpp
```diff
@@ -372,49 +375,79 @@ class GlobalDataHolder
...
     /* index */
-    using TIndexFile    = index_file<c_dbIndexType, c_origSbjAlph, c_redAlph>;
-    using TIndex        = typename TIndexFile::TIndex;
-    using TIndexCursor  = typename TIndex::cursor_type;
+    using TIndexFileFwd    = index_file<c_dbIndexType, c_origSbjAlph, c_redAlph, true>;
+    using TIndexFileRev    = index_file<c_dbIndexType, c_origSbjAlph, c_redAlph, false>;
+    using TIndexFwd        = typename TIndexFileFwd::TIndex;
+    using TIndexRev        = typename TIndexFileRev::TIndex;
+    using TIndexCursorFwd  = typename TIndexFwd::cursor_type;
+    using TIndexCursorRev  = typename TIndexRev::cursor_type;
```
h-2, line 417: You could make the BS-only types be `void` by default to state more clearly that they are only used in that specific mode.

src/shared_definitions.hpp
```diff
@@ -304,20 +311,21 @@ struct index_file_options
 /* actual index */
 template <DbIndexType           dbIndexType,
           AlphabetEnum          origAlph,
-          AlphabetEnum          redAlph>    // <- all members of index_file_options that influence types
+          AlphabetEnum          redAlph,
+          bool                  fwd>    // <- all members of index_file_options that influence types
 struct index_file
```
h-2, line 316: I don't think you should have two instances of `index_file`, instead you should have one `index_file` with two `index` members. The other things in the index should not be duplicated, right?

h-2 (commented):
You have adapted the naming scheme of renaming a lot of the current default things to `_fwd` and adding `_rev`. This is a little confusing since these terms are also used to refer to the two indeces inside the bidirectional index. I would prefer leaving the "default" things how they are named now and adding only a suffix to the new things, e.g, `_bs2` or `_bsrev` or something like that. It will also reduce the diff :)

sarahet:
Views tested with:

```
#include <iostream>

#include <seqan3/alphabet/nucleotide/all.hpp>
#include <seqan3/core/debug_stream.hpp>
#include "/project/hetzeldev/lambda3/lambda/src/view_translate_bisulfite.hpp"

using seqan3::operator""_dna4;

int main()
{
    seqan3::dna4_vector v1{"ACGTCGT"_dna4};

    auto out_fwd = v1 | seqan3::views::to_bisulfite_semialphabet(true);
    auto out_rev = v1 | seqan3::views::to_bisulfite_semialphabet(false);

    seqan3::debug_stream << v1 << '\n';

    for (auto i : out_fwd)
        seqan3::debug_stream << seqan3::to_rank(i);
    seqan3::debug_stream << '\n';

    for (auto i : out_rev)
        seqan3::debug_stream << seqan3::to_rank(i);
    seqan3::debug_stream << '\n';

    // Output:
    // ACGTCGT
    // 0121121
    // 3435435

    std::vector<std::vector<seqan3::dna4> > v2{"ACGTCGT"_dna4, "TCATGCA"_dna4};

    auto out_bs = v2 | seqan3::views::translate_bisulfite;
    seqan3::debug_stream << out_bs.size() << '\n'; // 4
    seqan3::debug_stream << out_bs[0].size() << '\n'; // 7

    for (auto i : out_bs)
    {
        for (auto j : i)
        {
            seqan3::debug_stream << seqan3::to_rank(j);
        }
        seqan3::debug_stream << '\n';
    }

    // Output
    // 0121121
    // 3435435
    // 1101210
    // 5435343
}

```

src/search_algo.hpp
```diff
@@ -533,19 +563,31 @@ search(LocalDataHolder<TGlobalHolder> & lH)
...
                 {
+                    bool const fwd = (subjNo % 2) - 1;
                     TMatch m {static_cast<typename TMatch::TQId>(i),
                               static_cast<typename TMatch::TSId>(subjNo),
                               static_cast<typename TMatch::TPos>(seedBegin),
                               static_cast<typename TMatch::TPos>(seedBegin + seedLength),
                               static_cast<typename TMatch::TPos>(subjOffset),
-                              static_cast<typename TMatch::TPos>(subjOffset + seedLength)};
+                              static_cast<typename TMatch::TPos>(subjOffset + seedLength),
+                              fwd};
```
h-2, line 573: Why do we still have these changes? Can't we compute `fwd` from `subjNo` (like we do above) wherever `fwd` is used?

src/search_algo.hpp
```diff
@@ -966,13 +1008,13 @@ iterateMatchesFullSimd(TLocalHolder & lH)
     for (auto it = lH.matches.begin(), itEnd = lH.matches.end(); it != itEnd; ++it)
     {
         // create blastmatch in list without copy or move
-        blastMatches.emplace_back(lH.qryIds [it->qryId / seqan::qNumFrames(TGlobalHolder::blastProgram)],
-                                  lH.gH.indexFile.ids[it->subjId / seqan::sNumFrames(TGlobalHolder::blastProgram)]);
+        blastMatches.emplace_back(lH.qryIds [it->qryId / TGlobalHolder::qryNumFrames],
+                                  lH.gH.indexFile.ids[it->subjId / TGlobalHolder::sbjNumFrames]);
```
h-2, line 1012: 👍🏻

src/search_algo.hpp
```diff
@@ -1158,14 +1200,15 @@ iterateMatchesFullSerial(TLocalHolder & lH)
     for (auto it = lH.matches.begin(), itEnd = lH.matches.end(); it != itEnd; ++it)
     {
         // create blastmatch in list without copy or move
-        record.matches.emplace_back(lH.qryIds [it->qryId / seqan::qNumFrames(TGlobalHolder::blastProgram)],
-                                    lH.gH.indexFile.ids[it->subjId / seqan::sNumFrames(TGlobalHolder::blastProgram)]);
+        bool fwd =  it->fwd;
```
h-2, line 1203: As I said, I think this can be computer here. Also maybe move it down to where it is used (l 1222)?.

src/search_datastructures.hpp
```diff
@@ -56,15 +56,17 @@ struct Match
     TPos subjStart;
     TPos subjEnd;
 
+    bool fwd = true;
```
h-2, line 59: I think this can go

src/search_datastructures.hpp
```diff
@@ -362,29 +364,11 @@ class GlobalDataHolder
...
-     seqan3::detail::lazy_conditional_t<c_transAlph == c_redAlph,
-                        decltype(seqan3::views::type_reduce(std::declval<TTransSbjSeqs &>())),   // no-op view
-                        seqan3::detail::lazy_conditional_t<c_transAlph != AlphabetEnum::AMINO_ACID,
-                                        seqan3::detail::lazy<TRedNuclAlphModString, TTransSbjSeqs, TRedAlph>,
-                                        seqan3::detail::lazy<TRedAlphModString, TTransSbjSeqs, TRedAlph> > >;
+    using TTransQrySeqs = typename TTransQrySeqsImpl<TQrySeqs, TOrigQryAlph, TTransAlph, TRedAlph>::type;
+    using TTransSbjSeqs = typename TTransSbjSeqsImpl<TSbjSeqs, TOrigSbjAlph, TTransAlph, TRedAlph>::type;
+
+    using TRedQrySeqs   = typename TRedSeqsImpl<TTransQrySeqs, TTransAlph, TRedAlph>::type;
+    using TRedSbjSeqs   = typename TRedSeqsImpl<TTransSbjSeqs, TTransAlph, TRedAlph>::type;
```
h-2, line 371: This is a very good change.

src/search_datastructures.hpp
```diff
@@ -423,6 +407,12 @@ class GlobalDataHolder
     using TBlastRepFile = seqan::FormattedFile<seqan::BlastReport, seqan::Output, TIOContext>;
     using TBamFile      = seqan::FormattedFile<seqan::Bam, seqan::Output, seqan::BlastTabular>;
 
+    // Number of frames for subject and query sequences
+    static constexpr uint8_t qryNumFrames = (c_redAlph == AlphabetEnum::DNA3BS ? 4 :
+                                 (seqan::qIsTranslated(blastProgram) ? 6 : (seqan::qHasRevComp(blastProgram) ? 2 : 1)));
```
h-2, line 412:
```suggestion
    static constexpr uint8_t qryNumFrames = (c_redAlph == AlphabetEnum::DNA3BS)  ? 4 :
                                            seqan::qIsTranslated(blastProgram)   ? 6 : 
                                            seqan::qHasRevComp(blastProgram)     ? 2 : 1;
```

The ternary operator binds in such a way that very legible chaining is possible :)

src/search_algo.hpp
```diff
@@ -430,7 +458,11 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
     // score the diagonal
     for (uint64_t i = 0; i < effectiveLength; ++i)
     {
-        s += lH.gH.scoringSchemePreScoring.score(qSeq[i], sSeq[i]);
+        if constexpr (fwd)
```
h-2, line 461: Can't you do the same thing here as in the alignment and get rid of the extra template parameter? `auto & scoringScheme = fwd ? ...` before the loop?

src/search_datastructures.hpp
```diff
@@ -437,8 +427,11 @@ class GlobalDataHolder
     TBlastRepFile       outfileBlastRep;
     TBamFile            outfileBam;
 
-    TScoreScheme3       scoringSchemePreScoring;
-    TScoreSchemeAlign   scoringSchemeAlign;
+    TScoreScheme3       scoringSchemePreScoringFwd;
+    TScoreScheme3       scoringSchemePreScoringRev;
+    TScoreSchemeAlign   scoringSchemeAlignFwd;
+    TScoreSchemeAlign   scoringSchemeAlignRev;
```
h-2, line 433: I kinda still prefer if we had `scoringSchemePreScoring` and `scoringSchemePreScoringBSRev` to clearly indicate that the second one is only used in the context of bisulphite stuff...

src/shared_definitions.hpp
```diff
@@ -259,17 +261,76 @@ using IndexSpec = seqan3::default_sdsl_index_type;
...
+{
+    using type = decltype(std::declval<TTransSeqs &>() | seqan3::views::dna_n_to_random<seqan3::dna4>);
+};
+
+template <typename TTransSeqs>
+struct TRedSeqsImpl<TTransSeqs, seqan3::dna5, seqan3::semialphabet_any<6>>
+{
+    using type = decltype(std::declval<TTransSeqs &>() | seqan3::views::dna_n_to_random<seqan3::dna4>
+                                                       | seqan3::views::translate_bisulfite);
+};
```
h-2, line 333: This solution is much better than the lazy-💩 we had before!

But maybe add one line of comment before everyone of these specialisations that documents the case?

src/view_pos_transform.hpp
```diff
@@ -0,0 +1,178 @@
...
+    const_reference operator[](size_type const n) const
+        requires seqan3::const_iterable_range<urng_t>
+    {
+        assert(n < size());
+        assert(pos_transform != nullptr);
+        return (*pos_transform)(urange, n);
+    }
+};
+
+
```
h-2, line 138:
```suggestion
```

src/view_pos_transform.hpp
```diff
@@ -0,0 +1,178 @@
...
+    static constexpr auto size_default_fn = [] (auto && urange) { return std::ranges::size(urange); };
+
+public:
+    template <typename pos_transform_t, typename size_transform_t = decltype(size_default_fn)>
+    constexpr auto operator()(pos_transform_t pos_transform, size_transform_t size_transform = size_default_fn) const
+    {
+        return seqan3::detail::adaptor_from_functor{*this, std::move(pos_transform), std::move(size_transform)};
+    }
+
+
```
h-2, line 154:
```suggestion
```

src/view_pos_transform.hpp
```diff
@@ -0,0 +1,178 @@
...
+//
+// Lambda is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+//
+// ==========================================================================
+// view_pos_transform.hpp
+// ==========================================================================
+
+#include <seqan3/range/views/all.hpp>
```
h-2, line 22: Please use the correct headers here and not `all.hpp`! I did that in my example because I was lazy, but it needlessly lengthens compile-time.

src/view_pos_transform.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+
```
h-2, line 24:
```suggestion
```

src/view_pos_transform.hpp
```diff
@@ -0,0 +1,178 @@
...
+        static_assert(std::ranges::sized_range<urng_t>,
+                      "The range parameter to views::pos_transform must model std::ranges::sized_range.");
+        static_assert(std::ranges::random_access_range<urng_t>,
+                      "The range parameter to views::pos_transform must model std::ranges::random_access_range.");
+
+        return view_pos_transform{std::forward<urng_t>(urange), std::move(pos_transform), std::move(size_transform)};
+    }
+};
+
+namespace seqan3::views
```
h-2, line 173:
```suggestion
namespace views
```

We shouldn't add things to SeqAn's namespace!

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+// Rank
+// 0 = A forward
+// 1 = C/T forward
+// 2 = G forward
+// 3 = A/G reverse
+// 4 = C reverse
+// 5 = T reverse
+
+static const std::array<uint8_t, 4> dna4_to_rank_bs_fwd = {0, 1, 2, 1};
+static const std::array<uint8_t, 4> dna4_to_rank_bs_rev = {3, 4, 3, 5};
```
h-2, line 47: make these `constexpr` instead of const and move them into the adaptor.

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+            }
+            else
+            {
+                return seqan3::assign_rank_to(dna4_to_rank_bs_rev[seqan3::to_rank(in)], seqan3::semialphabet_any<6>{});
+            }
+        });
+    }
+};
+
+namespace seqan3::views
```
h-2, line 82:
```suggestion
namespace views
```

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+//     }
+//
+//     template <std::ranges::range urng_t>
+//     constexpr friend auto operator|(urng_t && urange, translate_bisulfite_fn const & me)
+//     {
+//         return me(std::forward<urng_t>(urange));
+//     }
+// };
+
+struct translate_bisulfite_fn
```
h-2, line 248: Maybe 
```suggestion
struct reduce_to_bisulfite_fn
```
?

We are not doing translation and the number of frames doesn't increase in this step anymore.

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    template <std::ranges::range urng_t>
+    constexpr friend auto operator|(urng_t && urange, translate_bisulfite_fn const & me)
+    {
+        return me(std::forward<urng_t>(urange));
+    }
+};
+
+namespace seqan3::views
```
h-2, line 288:
```suggestion
namespace views
```

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+            }
+            else
+            {
+                return seqan3::complement(urange[size(urange) - pos - 1]);
+            }
+        });
+    }
+};
+
+namespace seqan3::views
```
h-2, line 328:
```suggestion
namespace views
```

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    template <std::ranges::range urng_t>
+    constexpr friend auto operator|(urng_t && urange, add_reverse_complement_fn const & me)
+    {
+        return me(std::forward<urng_t>(urange));
+    }
+};
+
+namespace seqan3::views
```
h-2, line 380:
```suggestion
namespace views
```

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    template <std::ranges::range urng_t>
+    constexpr friend auto operator|(urng_t && urange, duplicate_fn const & me)
+    {
+        return me(std::forward<urng_t>(urange));
+    }
+};
+
+namespace seqan3::views
```
h-2, line 423:
```suggestion
namespace views
```

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
+// ==========================================================================
```
h-2, line 1: We should make extra files for each view, I think.

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+                return seqan3::assign_rank_to(dna4_to_rank_bs_rev[seqan3::to_rank(in)], seqan3::semialphabet_any<6>{});
+            }
+        });
+    }
+};
+
+namespace seqan3::views
+{
+
+inline constexpr auto to_bisulfite_semialphabet = seqan3::views::deep{to_bisulfite_semialphabet_fn{}};
```
h-2, line 85: We don't need full Doxygen, but maybe a few comment lines per view what it does would be nice?

h-2 (commented):
Let's discuss optimising the views separately!

src/search.cpp
```diff
@@ -398,8 +398,8 @@ void realMain(LambdaOptions     const & options)
         #ifdef LAMBDA_MICRO_STATS
             localHolder.stats.timeSearch += sysTime() - buf;
         #endif
-
             // extend
+
             if (localHolder.matches.size() > 0)
```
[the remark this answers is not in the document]
sarahet: By accident :slightly_smiling_face:

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+        return std::forward<urng_t>(urange) | seqan3::views::pos_transform([] (auto && urange, size_t pos)
+        {
+            if (pos % 2 == 0)
+            {
+                return urange[(pos - (pos % 2)) / 2] | seqan3::views::reverse_complement_or_not(false);
+            }
+            else
+            {
+                return urange[(pos - (pos % 2)) / 2] | seqan3::views::reverse_complement_or_not(true);
+            }
```
h-2, line 369:
```suggestion
            return urange[pos / 2] | seqan3::views::reverse_complement_or_not(pos % 2 == 1);
```
?

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+                      "The inner range of the range parameter to views::translate_bisulfite must model "
+                      "std::ranges::random_access_range.");
+        static_assert(std::is_same_v<std::remove_const_t<seqan3::innermost_value_type_t<urng_t>>, seqan3::dna4>,
+                      "The range parameter to views::translate_bisulfite must be over a range over elements of "
+                      "seqan3::dna4.");
+
+        return std::forward<urng_t>(urange) | seqan3::views::pos_transform([] (auto && urange, size_t pos)
+        {
+            size_t index_strand = pos % 2;
+            return urange[pos] | seqan3::views::to_bisulfite_semialphabet(index_strand - 1); // We want forward strand first
```
h-2, line 277:
```suggestion
            return urange[pos] | seqan3::views::to_bisulfite_semialphabet(pos %2 == 1);
```
?

src/view_translate_bisulfite.hpp
```diff
@@ -0,0 +1,428 @@
...
+        return std::forward<urng_t>(urange) | std::views::transform([fwd] (auto const in)
+        {
+            if (fwd)
+            {
+                return seqan3::assign_rank_to(dna4_to_rank_bs_fwd[seqan3::to_rank(in)], seqan3::semialphabet_any<6>{});
+            }
+            else
+            {
+                return seqan3::assign_rank_to(dna4_to_rank_bs_rev[seqan3::to_rank(in)], seqan3::semialphabet_any<6>{});
+            }
```
h-2, line 77: Do the same thing here as for the other view with the function pointer.

src/view_translate_bisulfite.hpp
sarahet: Mine works as well, just not so nice :laughing:

src/view_pos_transform.hpp
```diff
@@ -0,0 +1,184 @@
...
+// The return type of views::pos_transform
+template <std::ranges::view urng_t, typename pos_transform_t, typename size_transform_t>
+class view_pos_transform : public ranges::view_base
+{
+
+private:
+    // The data members of view_pos_transform.
+    urng_t                              urange;
+    std::shared_ptr<pos_transform_t>    pos_transform = nullptr;
+    std::shared_ptr<size_transform_t>   size_transform = nullptr;
```
h-2, line 39: Can you replace `std::shared_ptr` with `seqan3::semiregular_box_t`? That might make the performance better.

src/view_reduce_to_bisulfite.hpp
```diff
@@ -0,0 +1,133 @@
...
+    // 4 = C reverse
+    // 5 = T reverse
+    static constexpr std::array<uint8_t, 4> dna4_to_rank_bs_fwd = {0, 1, 2, 1};
+    static constexpr std::array<uint8_t, 4> dna4_to_rank_bs_rev = {3, 4, 3, 5};
+
+    static auto func_fwd(auto && urange, size_t pos) { return seqan3::assign_rank_to(dna4_to_rank_bs_fwd[seqan3::to_rank(urange[pos])], seqan3::semialphabet_any<6>{}); }
+    static auto func_rev(auto && urange, size_t pos) { return seqan3::assign_rank_to(dna4_to_rank_bs_rev[seqan3::to_rank(urange[pos])], seqan3::semialphabet_any<6>{}); }
+
+    template <std::ranges::range urng_t>
+    constexpr auto operator()(urng_t && urange, bool const fwd) const
```
h-2, line 56:
```suggestion
    constexpr auto operator()(urng_t && urange, bsDirection const direction) const
```
?

h-2 (commented):
Just some minor stuff now.

src/view_dna_n_to_random.hpp
```diff
@@ -47,15 +49,15 @@ struct dna_n_to_random_fn
...
-            if (seqan3::to_char(in) == 'N')
+            if (seqan3::to_char(urange[pos]) == 'N')
             {
                 return seqan3::assign_rank_to(rng() % seqan3::alphabet_size<TAlph>, TAlph{});
             }
             else
             {
-                return static_cast<TAlph>(in);
+                return static_cast<TAlph>(urange[pos]);
             }
```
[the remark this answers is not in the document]
sarahet: The two lambdas trick did not work without the ```pos_transform``` for some reason ..

src/search_datastructures.hpp
```diff
@@ -423,6 +405,13 @@ class GlobalDataHolder
     using TBlastRepFile = seqan::FormattedFile<seqan::BlastReport, seqan::Output, TIOContext>;
     using TBamFile      = seqan::FormattedFile<seqan::Bam, seqan::Output, seqan::BlastTabular>;
 
+    // Number of frames for subject and query sequences
+    static constexpr uint8_t qryNumFrames = (c_redAlph == AlphabetEnum::DNA3BS)  ? 4 :
+                                            seqan::qIsTranslated(blastProgram)   ? 6 :
+                                            seqan::qHasRevComp(blastProgram)     ? 2 : 1;
+    static constexpr uint8_t sbjNumFrames = (c_redAlph == AlphabetEnum::DNA3BS)  ? 2 :
+                                            seqan::sIsTranslated(blastProgram)   ? 6 :
+                                            seqan::sHasRevComp(blastProgram)     ? 2 : 1;
```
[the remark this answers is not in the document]
sarahet: :stuck_out_tongue:

src/shared_definitions.hpp
```diff
@@ -228,7 +232,7 @@ using IndexSpec = seqan3::default_sdsl_index_type;
 //                                 16,
 //                                 10000000,
 //                                 sdsl::sa_order_sa_sampling<>,
-//                                 sdsl::isa_sampling<>,
+//                           TTransSbjSeqs      sdsl::isa_sampling<>,
```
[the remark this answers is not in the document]
sarahet: :eyes:

src/view_dna_n_to_random.hpp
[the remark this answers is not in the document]
sarahet: For some reason I could not combine it with the other views for some reason otherwise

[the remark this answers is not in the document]
h-2: Hm, that's strange and indicates some other problem. I can have a look after merging, but please don't mark such points as "resolved" 😉

h-2:
This LGTM now. However, the test on GCC8 is failing, because the link-time-optimiser crashes 😑 

That's definitely a bug in GCC, but we weren't triggering it before. Also, our test infrastructure should move to Github Actions ASAP.

Maybe you could look at the latter some time soon, and we deactivate the GCC8 test for now?

h-2 (approved):

PR #167, opened 2021-04-01
Small stuff

h-2:


src/view_pos_transform.hpp
```diff
@@ -34,13 +34,13 @@ class view_pos_transform : public ranges::view_base
...
-    std::shared_ptr<pos_transform_t>    pos_transform = nullptr;
-    std::shared_ptr<size_transform_t>   size_transform = nullptr;
+    urng_t                                        urange;
+    ::ranges::semiregular_box_t<pos_transform_t>  pos_transform;
+    ::ranges::semiregular_box_t<size_transform_t> size_transform;
 
     // Associated types iterator.
-    using reference         = decltype((*pos_transform)(urange, 0));
-    using const_reference   = decltype((*pos_transform)(std::as_const(urange), 0));
+    using reference         = decltype(pos_transform(urange, 0));
```
sarahet, line 42: I thought it needs to stay :wink:

src/mkindex.cpp
```diff
@@ -231,10 +231,10 @@ void realMain(LambdaIndexerOptions     const & options)
         if constexpr (c_transAlph == AlphabetEnum::AMINO_ACID)
             redSbjSeqs = transSbjSeqs | seqan3::views::deep{seqan3::views::convert<TRedSbjAlph>};
         else if constexpr (c_redAlph == AlphabetEnum::DNA3BS)
-            redSbjSeqs = transSbjSeqs | views::dna_n_to_random<seqan3::dna4>
+            redSbjSeqs = transSbjSeqs | views::dna_n_to_random
```
sarahet, line 234: Yes, that is much better

sarahet (approved):
LGTM

src/view_pos_transform.hpp
h-2: Yeah. So. semiregular_box is an addition to C++20. But only *exposion-only* which means that it's behaviour is specified and other components rely on it, but it's not actually there. The way it is specified is that it requires `operator*`, but the version that is in the range-v3 library does not require it 🤷🏻

PR #169, opened 2021-04-11
[MISC] put writing of record(s) into separate function

h-2:
This should make it easier to run iterate over matches of different directions in BS mode separately.


sarahet (approved):
LGTM

Very nice solution

PR #170, opened 2021-04-17
Fix bs untrue query and subject id and add bs tests

sarahet:
I added the bs tests separately so far because they don't go through SIMD mode currently. Later we can add them as separate program and only have one loop for all.


src/search_algo.hpp
```diff
@@ -598,7 +598,13 @@ _setFrames(TBlastMatch                          & bm,
...
+    } else if constexpr (TLocalHolder::TGlobalHolder::c_redAlph == AlphabetEnum::DNA3BS)
+    {
+        bm.qFrameShift = 1;
+        if (m.qryId % 4 > 1)
+            bm.qFrameShift = -bm.qFrameShift;
+    } else if constexpr (seqan::qHasRevComp(TLocalHolder::TGlobalHolder::blastProgram) &&
+                         TLocalHolder::TGlobalHolder::c_redAlph != AlphabetEnum::DNA3BS)
     {
         bm.qFrameShift = 1;
         if (m.qryId % 2)
```
h-2, line 610:
```suggestion
        if (m.qryId % TLocalHolder::TGlobalHolder::qNumFrames)
```

Shouldn't this be enough as a change (plus changing the if conditions)?

But in general, I don't understand why this check. Isn't every second frame a reverse complement? Why three out of four?

h-2 (commented):
I told you those IDs are tricky ^^

src/search_algo.hpp
```diff
@@ -598,7 +598,13 @@ _setFrames(TBlastMatch                          & bm,
         bm.qFrameShift = (m.qryId % 3) + 1;
         if (m.qryId % 6 > 2)
             bm.qFrameShift = -bm.qFrameShift;
-    } else if constexpr (seqan::qHasRevComp(TLocalHolder::TGlobalHolder::blastProgram))
+    } else if constexpr (TLocalHolder::TGlobalHolder::c_redAlph == AlphabetEnum::DNA3BS)
+    {
+        bm.qFrameShift = 1;
+        if (m.qryId % 4 > 1)
+            bm.qFrameShift = -bm.qFrameShift;
```
h-2, line 605:
```suggestion
        bm.qFrameShift = (m.qryId % 4 > 1) ? -1 : 1;
```

Should be easier to optimise.

src/search_algo.hpp
```diff
@@ -598,7 +598,13 @@ _setFrames(TBlastMatch                          & bm,
...
+    } else if constexpr (seqan::qHasRevComp(TLocalHolder::TGlobalHolder::blastProgram) &&
+                         TLocalHolder::TGlobalHolder::c_redAlph != AlphabetEnum::DNA3BS)
```
h-2, line 607:
```suggestion
    } else if constexpr (seqan::qHasRevComp(TLocalHolder::TGlobalHolder::blastProgram))
```

Not required because if-else?

src/search_algo.hpp
```diff
@@ -613,7 +619,8 @@ _setFrames(TBlastMatch                          & bm,
         bm.sFrameShift = (m.subjId % 3) + 1;
         if (m.subjId % 6 > 2)
             bm.sFrameShift = -bm.sFrameShift;
-    } else if constexpr (seqan::sHasRevComp(TLocalHolder::TGlobalHolder::blastProgram))
+    } else if constexpr (seqan::sHasRevComp(TLocalHolder::TGlobalHolder::blastProgram) ||
+                         TLocalHolder::TGlobalHolder::c_redAlph == AlphabetEnum::DNA3BS)
```
h-2, line 623: Are you sure this change is correct? We have two frames on the subject, but there is never revcomp on the subject, is there?

src/search_algo.hpp
h-2: Anyway, this means

0 -> frameshift 1
1 -> frameshift 1
2 -> frameshift -1
3 -> frameshift -1

h-2: So frameshift actually implies "reverse comlpement". The information for "second translation frame" is lost. Or is it the other way around?

src/search_algo.hpp
h-2: here we necode sFrameShift to mean second "translation" frame and not revcomp, or not?

src/search_algo.hpp
sarahet: Yes, I thought this is what the frameshift implies for the normal nucleotide mode as well. Please correct me if that was wrong.

src/search_algo.hpp
sarahet: True ..

h-2 (approved):

h-2:
It all LGTM!

PR #171, opened 2021-05-16
[Test] Update search test calls and md5sums

sarahet:


sarahet:
We can't move to the SeqAn3 test interface yet as included in the App repo, the SeqAn3 version is too old

h-2:
Changing the options doesn't change the results? 🤔

h-2 (approved):
Sorry, I didn't see that the .gz file was changed 🙈

PR #172, opened 2021-05-21
Add bisulfite SIMD mode

sarahet:
Not sure this is ideal or what you had in mind but I thought it would be a not too complicated solution without many overloads etc


sarahet:
I accidentally included the other commits as well here from the old PR. But maybe after we discussed and improved this one here, I can just delete the old PR and have all in one here plus SIMD tests.

h-2:
Yes, that looks like a quite simple solution. Everything works like this?

sarahet:
You mean from the other modes?

sarahet:
GCC9 on linux fails, I also saw that already on our server but didn't have time to check and see what the problem was. Also on our servers it was different to the issue here (we had a problem I think in the linking step)

src/search_algo.hpp
```diff
@@ -933,7 +934,8 @@ _performAlignment(TDepSetH & depSetH,
     unsigned constexpr sizeBatch = seqan::LENGTH<TSimdAlign>::VALUE;
     unsigned const      fullSize = sizeBatch * ((seqan::length(blastMatches) + sizeBatch - 1) / sizeBatch);
 
-    TSimdScore simdScoringScheme(lH.gH.scoringSchemeAlign);
+    auto currentScoringScheme = dir == bsDirection::fwd ? lH.gH.scoringSchemeAlign : lH.gH.scoringSchemeAlignBSRev;
```
h-2, line 937:
```suggestion
    auto const & currentScoringScheme = dir == bsDirection::fwd ? lH.gH.scoringSchemeAlign : lH.gH.scoringSchemeAlignBSRev;
```

💅🏻

src/search_algo.hpp
```diff
@@ -994,6 +996,13 @@ iterateMatchesFullSimd(TLocalHolder & lH)
     std::list<TBlastMatch> blastMatches;
     for (auto it = lH.matches.begin(), itEnd = lH.matches.end(); it != itEnd; ++it)
     {
+        if constexpr (TLocalHolder::TGlobalHolder::c_redAlph == AlphabetEnum::DNA3BS)
```
h-2, line 999:
```suggestion
        // In BS-mode, skip those results that have wrong orientation
        if constexpr (TLocalHolder::TGlobalHolder::c_redAlph == AlphabetEnum::DNA3BS)
```

src/search_algo.hpp
```diff
@@ -1051,7 +1060,9 @@ iterateMatchesFullSimd(TLocalHolder & lH)
     _setupDepSets(depSetH, depSetV, blastMatches);
 
     // Run extensions WITHOUT ALIGNMENT
-    _performAlignment(depSetH, depSetV, blastMatches, lH, std::false_type());
+    _performAlignment(depSetH, depSetV, blastMatches, lH, std::false_type(), dir);
+
+    auto currentScoringScheme = dir == bsDirection::fwd ? lH.gH.scoringSchemeAlign : lH.gH.scoringSchemeAlignBSRev;
```
h-2, line 1065:
```suggestion
    auto const & currentScoringScheme = dir == bsDirection::fwd ? lH.gH.scoringSchemeAlign : lH.gH.scoringSchemeAlignBSRev;
```

h-2 (commented):
We need to figure out the test failure, otherwise it looks good!

sarahet:
Tests fixed themselves :sweat_smile:

src/view_pos_transform.hpp
```diff
@@ -30,7 +30,7 @@
 
 // The return type of views::pos_transform
 template <std::ranges::view urng_t, typename pos_transform_t, typename size_transform_t>
-class view_pos_transform : public std::ranges::view_interface<view_pos_transform<urng_t, pos_transform_t, size_transform_t>> //public ranges::view_base
+class view_pos_transform : public ranges::view_base
```
h-2, line 33:
```suggestion
class view_pos_transform : public std::ranges::view_base
```

?

PR #174, opened 2021-10-26
Update to SeqAn 3.1.0 pre-release

sarahet:


src/mkindex_misc.hpp
```diff
@@ -87,7 +85,7 @@ _readMappingFileUniProt(TInputView                                            &
             taxIdIsPresent[idNum] = true;
         }
 
-        seqan3::detail::consume(fiv | seqan3::detail::take_line);
+        seqan3::detail::consume(fiv | std::views::take_while([](auto && chr){ return chr != '\n'; }));
```
h-2, line 88: This still needs to check for `== '\r'`.

You can also add this somewhere globally:
```cpp
inline constexpr auto not_eol = !(seqan3::is_char<'\r'> || seqan3::is_char<'\n'>);
```
And then just use `std::views::take_while(not_eol)` in all these places

src/mkindex_misc.hpp
```diff
@@ -100,26 +98,26 @@ _readMappingFileNCBI(TInputView                                            & fiv
                      std::unordered_map<std::string, uint64_t>       const & accToIdRank)
 {
     // skip line with headers
-    seqan3::detail::consume(fiv | seqan3::detail::take_line_or_throw);
+    seqan3::detail::consume(fiv | std::views::take_while([](auto && chr){ return chr != '\n'; }));
```
h-2, line 101: here as well

src/mkindex_misc.hpp
sarahet: Yes, sorry I missed that ..

h-2 (approved):

PR #176, opened 2022-06-14
Use fmindex_collection as an FMIndex

SGSSGene:
This replaces [SeqAn3](https://github.com/seqan/seqan3/) based data structures and search routines with those from [fmindex_collection](https://github.com/SGSSGene/fmindex-collection)

fmindex_collection also has an integrated version of libsais which is "faster" than libdivsufsort (which is integrated in sdsl/SeqAn3)


src/mkindex_algo.hpp
```diff
@@ -630,34 +635,34 @@ auto parseAndStoreTaxTree(std::vector<bool>          & taxIdIsPresent,
...
 {
     using TRedAlph       = seqan3::range_innermost_value_t<TStringSet>;
+    using TIndexSpec     = IndexSpec<index_t, seqan3::alphabet_size<TRedAlph>>;
+    using TIndex         = typename TIndexSpec::index_type;
 
-    constexpr auto is_collection = seqan3::text_layout::collection;
-    using TSpec = IndexSpec<seqan3::alphabet_size<TRedAlph>>;
-    using TIndex = std::conditional_t<is_bi,
-                                      seqan3::bi_fm_index<TRedAlph, is_collection, TSpec>,
-                                      seqan3::fm_index<TRedAlph, is_collection, TSpec>>;
```
h-2, line 644: same here, I don't think this needs to be changed since we only have two index types

src/mkindex_options.hpp
```diff
@@ -187,10 +187,12 @@ void parseCommandLine(LambdaIndexerOptions & options, int argc, char const ** ar
...
 
     // set db index type
-    if (dbIndexTypeTmp == "bifm")
+    if (dbIndexTypeTmp == "fm")
+        options.indexFileOptions.indexType = DbIndexType::FM_INDEX;
+    else if (dbIndexTypeTmp == "bifm")
         options.indexFileOptions.indexType = DbIndexType::BI_FM_INDEX;
     else
-        options.indexFileOptions.indexType = DbIndexType::FM_INDEX;
+        throw seqan3::argument_parser_error("ERROR: Unknown index type \"" + dbIndexTypeTmp + "\"");
```
h-2, line 195: argument parser already checks the allowed values

src/search_algo.hpp
```diff
@@ -400,47 +406,100 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
...
-                                }};
-
-    seqan3::search(seed, lH.gH.indexFile.index, cfg);
+
+    assert(lH.queries.size() == 1);
+    assert(lH.queries[0].size() == seed.size());
+
+    if constexpr (c_indexType == DbIndexType::FM_INDEX)
+    {
+
```
h-2, line 421:
```suggestion
```

src/search_algo.hpp
```diff
@@ -400,47 +406,100 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
...
+        // prepare query for fmindex_collection, by shifting by one and reversing the string
+        for (size_t i{0}; i < seed.size(); ++i)
+        {
+            lH.queries[0][seed.size()-i-1] = seed[i].to_rank()+1;
+        }
```
h-2, line 426: Do we need to copy all the seeds into rank-versions? Why does a view not work?

src/search_algo.hpp
```diff
@@ -400,47 +406,100 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
...
+            lH.gH.indexFile.index,
+            lH.queries,
+            lH.options.maxSeedDist,
+            [&](size_t /*queryId*/, auto cursor, size_t /*errors*/)
+            {
+                lH.cursor_buffer.push_back(cursor);
+            }
+        );
+    }
+
```
h-2, line 438:
```suggestion
```

src/search_algo.hpp
```diff
@@ -400,47 +406,100 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
...
+            }
+        );
+    }
+    else
+    {
+        []<bool flag = false>()
+        {
+            static_assert(flag, "unsupported DbIndexType");
+        };
+    }
```
h-2, line 463: See above, I think we shouldn't need to perform these checks everywhere.

src/search_algo.hpp
```diff
@@ -400,47 +406,100 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
...
 
-template <typename TGlobalHolder, typename TSeed>
+template <DbIndexType   c_indexType,
+          typename      TGlobalHolder,
+          typename      TSeed>
 inline void
 searchHalfExactImpl(LocalDataHolder<TGlobalHolder> & lH, TSeed && seed)
 {
+
+    auto extendRight = [&](auto& cursor, auto c) {
```
h-2, line 473:
```suggestion
    auto extendRight = [&](auto & cursor, auto c)
    {
```

h-2: formatting in other places as well

src/search_algo.hpp
```diff
@@ -455,14 +514,14 @@ searchHalfExactImpl(LocalDataHolder<TGlobalHolder> & lH, TSeed && seed)
                     alph_t cur_letter = seqan3::assign_rank_to(r, alph_t{});
 
                     lH.cursor_tmp_buffer2.emplace_back(cursor, error_count + (cur_letter != seed_at_i));
-                    if (!lH.cursor_tmp_buffer2.back().first.extend_right(cur_letter))
+                    if (!extendRight(lH.cursor_tmp_buffer2.back().first, cur_letter))
```
h-2, line 517: There is member `.extendRight()` above, but here you switc member for free function, why?

src/search_algo.hpp
```diff
@@ -475,7 +534,9 @@ searchHalfExactImpl(LocalDataHolder<TGlobalHolder> & lH, TSeed && seed)
     lH.cursor_tmp_buffer.clear();
 }
 
-template <typename TGlobalHolder>
+
+template <DbIndexType   c_indexType,
+          typename      TGlobalHolder>
```
h-2, line 539: indexType is already in the globalholder, shouldn't be an extra template arg

src/search_algo.hpp
```diff
@@ -568,25 +629,24 @@ search(LocalDataHolder<TGlobalHolder> & lH)
...
                         {
                             // revert last extension
-                            cursor = old_cursor;
                             break;
                         }
-
                         ++seedLength;
+                        cursor = newCursor;
                         old_count = new_count;
-                        old_cursor = cursor;
```
h-2, line 589: Are the changes in this section semantic or did you just switch "old <-> current" to "current <-> new" ? It is a little bit confusing, because there is also old_count. 

I don't think it was ideal before, but I am not sure it is better now 😕

src/search_algo.hpp
```diff
@@ -595,8 +655,11 @@ search(LocalDataHolder<TGlobalHolder> & lH)
                     continue;
 
                 // locate hits
-                for (auto [ subjNo, subjOffset ] : cursor.lazy_locate())
-                {
+                for (auto [subjNo, subjOffset] : fmindex_collection::LocateLinear{lH.gH.indexFile.index, cursor}) {
+                    // !TODO Should this be handled by the cursor?
+                    if (c_indexType == DbIndexType::FM_INDEX) {
+                        subjOffset -= seedLength;
```
h-2, line 661: The index returns the end of the match?

src/search_datastructures.hpp
```diff
@@ -505,14 +509,20 @@ class LocalDataHolder
     // regarding the gathering of stats
     StatsHolder         stats{};
 
+    // search work space
+    std::vector<std::vector<uint8_t>> queries;
```
h-2, line 513: Like I said, ideally we would use views instead. If we cannot, I would propose to use `TCDStringSet<std::vector<uint8_t>>` instead, because it avoid dynamic reallocations (just one buffer that is reused).

src/shared_definitions.hpp
```diff
@@ -48,16 +56,16 @@
...
 
 inline std::string
 _indexEnumToName(DbIndexType const t)
 {
     switch (t)
     {
-        case DbIndexType::FM_INDEX:      return "fm_index";
-        case DbIndexType::BI_FM_INDEX:   return "bi_fm_index";
+        case DbIndexType::FM_INDEX:        return "fm_index";
+        case DbIndexType::BI_FM_INDEX:     return "bi_fm_index";
```
h-2, line 68: undo whitespace changes, also the comman in line 59

src/search_algo.hpp
SGSSGene: I will revert this to be closer to the original.

src/search_algo.hpp
SGSSGene: Yes, since we are using the reversed text to be able to extend our query to the right, the suffix array reports the ending of our text, not the beginning. To be able to determine the start position the cursor would need to carry the information how often it was extended....
But, I think I should change this. Just haven't got to it yet and seemed to have a low priority.

src/shared_definitions.hpp
SGSSGene: Mhm...Maybe I havn't figured it out. There are 41 characters before the `return ...` should be 40 or 44?

src/search_algo.hpp
SGSSGene: True!

src/search_datastructures.hpp
SGSSGene: I am assuming that LocalDataHolder is instantiated once per thread (I hope this is true, otherwise I broke the code).
The search function currently takes a std::vector<std::vector<uint8_t>>. It is also only allocated once when creating LocalDataHolder. This could be removed if I switch over to ranges in my library.

src/mkindex_options.hpp
SGSSGene: I just noticed, this is not true. `index_file_options::indexType` has no default initialized value, I put it there....no clue what I was thinking. :thinking:

src/mkindex_algo.hpp
SGSSGene: reverted (adjusted a little, so it is closer to the code in shared_definitions.hpp)

src/search_algo.hpp
h-2: You are always assiging to `lH.queries[0]`, so why is there a vector-of-vector at all?

src/search_datastructures.hpp
h-2: Yeah, I would really prefer if we could get rid of this. Is it feasible time-wise to change this in your library? Like I said, you could do some very simple checks for begin() and end() without going all crazy with concepts...

If it is not feasible, I would still prefer if we can rename this to `queries_as_ranks` or something like that. And also attach a `TODO remove this as soon as fmindex_collection can handle ranges`

h-2 (changes requested):

src/search_algo.hpp
```diff
@@ -572,8 +626,7 @@ search(LocalDataHolder<TGlobalHolder> & lH)
                     size_t old_count = cursor.count();
                     while (seedBegin + seedLength < lH.redQrySeqs[i].size())
                     {
-                        if (!cursor.extend_right(lH.redQrySeqs[i][seedBegin + seedLength]))
-                            break;
+                        cursor = cursor.extendRight((lH.redQrySeqs[i][seedBegin + seedLength]).to_rank()+1);
```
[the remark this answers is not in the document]
SGSSGene: No, is not required. The state "invalid" doesn't really exists (even in SeqAn3), it is the same as an empty cursor.

In SeqAn3 case, if an cursor was empty, it was rolled backed, and "invalid" was returned.
In my case I don't unroll and just assume the user will check for `count()` any ways (or doesn't, which is also ok).

src/search_algo.hpp
```diff
@@ -595,8 +648,13 @@ search(LocalDataHolder<TGlobalHolder> & lH)
                     continue;
 
                 // locate hits
-                for (auto [ subjNo, subjOffset ] : cursor.lazy_locate())
+                for (auto [subjNo, subjOffset] : fmindex_collection::LocateLinear{lH.gH.indexFile.index, cursor})
```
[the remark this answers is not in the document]
SGSSGene: No, it doesn't, There should be no allocation involved besides the index construction/loading.

src/mkindex_algo.hpp
```diff
@@ -636,28 +642,16 @@ auto generateIndex(TStringSet                       & seqs,
...
-        // WORKAROUND https://github.com/seqan/seqan3/pull/1519
-        std::vector<std::vector<TRedAlph>> tmp = seqs | seqan3::views::to<std::vector<std::vector<TRedAlph>>>;
-        index = TIndex{tmp};
-    }
-    else
-    {
-        index = TIndex{seqs};
-    }
+    TIndex index{seqs | std::views::transform([] (auto const& seq) { return seq | seqan3::views::to_rank; })
+            | fmindex_collection::add_sentinels, 16};
```
sarahet, line 654: Is this the sampling rate?

SGSSGene: Yes, that is the sampling rate.

src/mkindex_algo.hpp
```diff
@@ -636,28 +642,17 @@ auto generateIndex(TStringSet                       & seqs,
...
-        // WORKAROUND https://github.com/seqan/seqan3/pull/1519
-        std::vector<std::vector<TRedAlph>> tmp = seqs | seqan3::views::to<std::vector<std::vector<TRedAlph>>>;
-        index = TIndex{tmp};
-    }
-    else
-    {
-        index = TIndex{seqs};
-    }
+    // creates the index with sampling rate of 5
+    TIndex index{seqs | std::views::transform([] (auto const& seq) { return seq | seqan3::views::to_rank; })
```
h-2, line 654:
```suggestion
    TIndex index{seqs | seqan3::views::to_rank
```

This should be enough, because the view is a "deep" view.

h-2 (commented):
one minor thing left!

h-2:
Why do all the tests fail? Shouldn't just the indexer tests fail?

I assume the results for the search are the same as before, or are they a little different now?

sarahet:
> Why do all the tests fail? Shouldn't just the indexer tests fail?
> 
> I assume the results for the search are the same as before, or are they a little different now?

They are also different because Simon said there are some edge cases that are not considered by the sdsl I think. Need to search through the gitter history though, it’s been a while.

sarahet:
Also aren’t our index files not also tested in the search tests? 🤔

h-2:
good point. maybe we should change that.

sarahet:
We anyways thought of updating the tests to move to a SeqAn3-like test structure, right? I could work on the test update in general and split this plus add the new index/search output files

h-2:
> We anyways thought of updating the tests to move to a SeqAn3-like test structure, right?

I wouldn't mind if you want to do this, but it is not a high prioirity for me.

>  I could work on the test update in general and split this plus add the new index/search output files

Yeah, it would be good if the search test don't also test the index.

@SGSSGene Can you do the last little change and then we can get this in?

src/search_algo.hpp
```diff
@@ -400,47 +406,77 @@ seedLooksPromising(LocalDataHolder<TGlobalHolder> const & lH,
     return false;
 }
 
-template <typename TGlobalHolder, typename TSeed>
+template <typename      TGlobalHolder,
+          typename      TSeed>
```
h-2, line 410: revert NOP

P.S: I really need to add clang-format 🙈

src/search_algo.hpp
```diff
@@ -475,6 +511,7 @@ searchHalfExactImpl(LocalDataHolder<TGlobalHolder> & lH, TSeed && seed)
     lH.cursor_tmp_buffer.clear();
 }
 
+
```
h-2, line 514:
```suggestion
```

h-2 (approved):
just minor 💅🏻

src/mkindex_algo.hpp
```diff
@@ -636,28 +642,15 @@ auto generateIndex(TStringSet                       & seqs,
...
-
-    constexpr auto is_collection = seqan3::text_layout::collection;
-    using TSpec = IndexSpec<seqan3::alphabet_size<TRedAlph>>;
-    using TIndex = std::conditional_t<is_bi,
-                                      seqan3::bi_fm_index<TRedAlph, is_collection, TSpec>,
-                                      seqan3::fm_index<TRedAlph, is_collection, TSpec>>;
+    using TIndexSpec     = IndexSpec<seqan3::alphabet_size<TRedAlph>>;
+    using TIndex         = std::conditional_t<is_bi,
+                                             fmindex_collection::BiFMIndex<TIndexSpec>,
+                                             fmindex_collection::ReverseFMIndex<TIndexSpec>>;
```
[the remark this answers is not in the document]
SGSSGene: So `FMIndex` is the normal FMIndex which allows backward searches. A cursor always points beginning of a search query.

`ReverseFMIndex` reverses the text before building the index. This means it will allow only forward searches, with the downside that internally cursors now point to the end of the pattern. (This is compensated by carrying an extra cursor depth/query length variable to correct for this).

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "submodules/seqan3"]
 	path = submodules/seqan3
 	url = ../../seqan/seqan3.git
+[submodule "submodules/fmindex-collection"]
+	path = submodules/fmindex-collection
+	url = https://github.com/SGSSGene/fmindex-collection.git
```
h-2, line 9: ``suggestion
	url = ../../fmindex-collection.git
```
If this is not your final version, can you change this?

File: .gitmodules [hunk 22]
```diff
@@ -4,3 +4,6 @@
 [submodule "submodules/seqan3"]
 	path = submodules/seqan3
 	url = ../../seqan/seqan3.git
+[submodule "submodules/fmindex-collection"]
+	path = submodules/fmindex-collection
+	url = ../../fmindex-collection.git
```
h-2, line 9:
```suggestion
	url = ../../SGSSGene/fmindex-collection.git
```

h-2: Sorry, I had this wrong initially. I immediately corrected in my comment, but it seems you were too fast in copying the old code.

h-2 (approved):
This only needs to be rebased and stashed now, I think.

PR #177, opened 2022-07-08
[fix] keep match with highest score

SGSSGene:
Fixes the sorting of matches.
Before lambda3 sorted the matches and outputted the first entry, which wasn't always the match with the highest score.

This additionally sorts by the inverted bitScore to ensure that `unique` keeps the match with the highest score.


h-2:
Thanks for finding this issue 🎉 

I discussed a different solution with @sarahet, she will contact you about it!

SGSSGene:
Here an example from the unittest data, what changes in the results. This shows `sdiff -s output.original.m8 output.new.m8`.
```
qry214  ref764  86.89   122     12      3       1       118   | qry214  ref764  85.71   126     6       4       1       118
                                                              > qry282  ref771  70.34   118     26      3       17      130
qry282  ref771  70.43   115     31      3       17      130   <
                                                              > qry973  ref849  91.54   130     9       2       1       129
                                                              > qry973  ref931  91.06   123     9       2       1       122
qry973  ref849  76.47   136     18      7       1       129   <
qry973  ref931  74.24   132     14      7       1       122   <
```

h-2 (approved):

PR #178, opened 2022-07-08
Feat/heapless halfexact backtracking

SGSSGene:
This is based on (and blocked by) #176 

It changes the halfExact search by removing the heap structures and replacing it with recursive function calls.

This still needs performance evaluation, to make sure it is not slower than before.

### DIFF
--- a/src/search_algo.hpp
+++ b/src/search_algo.hpp
@@ -452,6 +452,50 @@ searchHalfExactImpl(LocalDataHolder<TGlobalHolder> & lH, TSeed && seed)
 
     // Function which implements naive backtracking
     std::function<void(TIndexCursor, size_t, size_t)> backtrackingStep;
+    #if 0 // if activated, this only does a recursive call on an error
+    backtrackingStep = [&backtrackingStep, &lH, &seed](TIndexCursor cursor, size_t errorsLeft, size_t i)
+    {
+        // check if end of seed is reached (report as results)
+        if (i == lH.options.seedLength)
+        {
+            lH.cursor_buffer.emplace_back(cursor);
+            return;
+        }
+
+        // No errors allowed? performe exact search
+        if (errorsLeft == 0)
+        {
+            for (;i < lH.options.seedLength; ++i)
+            {
+                cursor = cursor.extendRight(seed[i].to_rank() + 1);
+                if (cursor.empty())
+                    return;
+            }
+            lH.cursor_buffer.emplace_back(cursor);
+        }
+        else
+        {
+            // only branches on error, stays iterative on no errors
+            for (; i < lH.options.seedLength and !cursor.empty(); ++i)
+            {
+                auto curRank    = seed[i].to_rank() + 1ul;
+                auto newCursors = cursor.extendRight();
+                for (size_t r{1}; r < newCursors.size(); ++r)
+                {
+                    if (!newCursors[r].empty() and curRank != r)
+                    {
+                        backtrackingStep(newCursors[r], errorsLeft - 1, i+1);
+                    }
+                }
+                cursor = newCursors[curRank];
+            }
+            if (!cursor.empty())
+            {
+                lH.cursor_buffer.emplace_back(cursor);
+            }
+        }
+    };
+    #else // else case, does a recursion on every step (original PR had this)
     backtrackingStep = [&backtrackingStep, &lH, &seed](TIndexCursor cursor, size_t errorsLeft, size_t i)
     {
         // check if end of seed is reached (report as results)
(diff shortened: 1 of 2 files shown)


h-2:
I don't think this will improve performance. Recursive function calls cannot be inlined, and in long proteins we can reach maximum recursion depth (had this problem in Lambda2 where it is implemented as a recursion). The guidance I always received was to avoid recursion unless you can prove that tail-call optimisation happens (which alleviates the performance problems of recursion).

Also, there shouldn't be any dynamic memory allocations in the current code on an average iteration (`clear()` does not free).

But of course you can convince me of the opposite, if it shows to be faster :)

SGSSGene:
Yes, I am not expecting a speed improvement by removing the heap access (and those few allocation hapepning shouldn't matter at all). I am not even sure if it might even hurt performance, because of how it traverses the FMIndex:
Citing my self: "This still needs performance evaluation, to make sure it is not slower than before." :stuck_out_tongue: 

Let me describe briefly my motivation for this PR:
- I had it written anyways, to find the "bug" from #177, it is easier for me to reason and step through the recursive version for debugging reasons.
- It removes the extra data structure in LocalDataHolder, I thought you might like that :smile: 
- It is adjusted for the cursor of fmindex_collection and its semantics of `extendRight`
- It uses a more performant version of `extendRight()` for the case of substitution error. (not really expecting a speed improvement from this)

Your point about the recursion is interesting. I'd assume that the recursion uses almost no stack:
parameters:
  32 bytes for the cursor (bidirectional index cursor is 40bytes)
    8 bytes for errosLeft
    8 bytes for i
  32 bytes for newCursor before calling the recursion (This hopefully can be optimized away with the parameter cursor, but not sure)
+ some bytes for the recursive call (how much is needed here? 8bytes should enough, shouldn't it?)
So that is 80 bytes per recursion call. (I should measure this!)


I need a recursion call per base pair. Which means I could do 10'000 steps which would require 800kbytes. Wouldn't this fit easily into my stack? (seqan3 requires for fm index 50bytes per cursor, and 66bytes for the bifm). 

So I am surprised that you had recursion issues. How long are the expected halfExact searches?

My implementations in `fmindex_collection::search_xyz()` uses a similar recursive call, so I am really invested in this, wanting to make sure it is the fastest and smallest :+1: 

I would now go forward as follows: Wait till #177 is resolved in someway. Wait till #176 is merged. Then I would take some measurements about performance of this PR. I will also check how much the stack grows and what the maximum recursion depth is. And then I will have learned something, we can evaluate the results and I will try to convince you again ;-)

h-2:
> it is easier for me to reason and step through the recursive version for debugging.

Yeah, I agree with this completely. It took me a while to create the non-recursive version, because that's not the way we are taught these algorithms.

> I'd assume that the recursion uses almost no stack

I am by no means an expert on these things, but AFAIK there is a non-negligible overhead to just invoking a non-inlinable function. This is independent of the actual stack size (which contains more than the local variables though), and is more expensive than jumping to the beginning of the loop.

As I said, I reached maximum recursion depth in lambda2. This is actually quite trivial to reproduce: just search a database in itself. if you have a sequence that is identical in input and output, but whose seeds appear almost nowhere else, it will lead to extending the first seed to the full sequence length, because doing so does not reduce the number of hits. Depending on the sequence length, you could end up with thousands of recursion steps.

> My implementations in fmindex_collection::search_xyz() uses a similar recursive call, so I am really invested in this, wanting to make sure it is the fastest and smallest

Yeah, I understand that. I would also recommend trying to benchmark this outside of lambda to get better estimates on how this influences performance. But as it stands right now, you would have show very convinving numbers for me to change this 😄

SGSSGene:
TL DR: Your implementation is the fastest, I wouldn't accept this PR.

So after looking at the backtracking algorithm, I noticed, I already implemented a version that only branches if an error occurse, keeping recursions to the maximum of allowed numbers of errors. So I also implemented this for `half-exact` searches and added it to this PR, but it turns out your implementation is the fastest. I ran it with some data 5 times and took the best run for comparison:

```
half-exact: fastest     runs
orig:       16.4s       16.6s/16.5s/16.6s/16.6s/16.4s  # original implementation
recursive:  17.6s       18.3s/18.0s/18.0s/18.0s/17.6s  # this PR
error-rec:  16.8s       17.3s/16.7s/16.8s/17.3s/17.4s  # this PR - using "recursion on error"
```

call: `./bin/lambda3 searchn -q illumina_reads_100.fasta -v 2 --adaptive-seeding false --seed-length 50 --seed-delta 2 --seed-half-exact 1 --threads 2 -i fm.index.lba`

h-2:
OK, then I will close this now.

PR #179, opened 2022-07-13
Pre search filter

sarahet:
Introduces a pre-search filter for the nucleotide mode that checks whether seeds only contain a single letter.


src/search_algo.hpp
```diff
@@ -529,6 +529,23 @@ search(LocalDataHolder<TGlobalHolder> & lH)
...
+            {
+                bool all_equal = true;
+                for (size_t j = seedBegin + 1; j < (seedBegin + lH.options.seedLength); j++)
+                {
+                    if (lH.redQrySeqs[i][j] != lH.redQrySeqs[i][seedBegin])
+                    {
+                        all_equal = false;
+                        break;
+                    }
+                }
```
h-2, line 543: This can be made a bit more readable with std::equal() or std::adjacent_find, but that's 💅🏻

h-2 (commented):
In general, would it maybe be a simpler (and also faster?) solution to check if the current letter equals the next and if yes `++seedBegin`? 

It could be integrated this with the ` // skip proteine 'X' or Dna 'N'` ...

What do you think?

Also, I think we can have this for protein search, too.

sarahet:
The tests fail because of something related to the dna_n_to_random view (very weird).

Regarding your suggestion: But wouldn't we want to go forward with a seed that e.g. looks like AAGCTAGCA and not move the seedBegin in this case?

h-2:
> The tests fail because of something related to the dna_n_to_random view (very weird).

Can you get it fixed, or shall I have a look?

> Regarding your suggestion: But wouldn't we want to go forward with a seed that e.g. looks like AAGCTAGCA and not move the seedBegin in this case?

Why not? I assume that fewer repetitions always increase the usefulness of a seed. Although, this approach will only prevent AAAAAAAT and not TAAAAAAAA.

The only situation where it would be detrimental is, if it is the last possible seed in a sequence, and we don't search at all, because we extend. I am not sure how big this effect would be, though.

sarahet:
New solution implemented

PR #180, opened 2022-07-14
update to current SeqAn3 master and add compiler workarounds

h-2:
This fixes problems with newer versions of GCC11 and fixes *some* of the problems with GCC10.4, but that last remaining problem I have been unable to fix in the last couple of hours. We will see this on Linux as well as soon as that gets GCC10.4 :|

@SGSSGene do you like weird compiler problems and want to give it a try? 😉 

In any case, I think we can merge these changes (although they are also workarounds at best).


sarahet (approved):
Thanks for digging into this mess!

PR #181, opened 2022-07-15
[fix] GCC10.4 build and simplify size computation

h-2:


h-2:
@SGSSGene @sarahet This fixes the builds on GCC10.4 and is a bit easier to understand than before. There is a bug (we should be using the exception case in more situations), but fixing it would change the evalues of many files and thereby also the checksums. 

If the other things don't get worked on over the weekend, we can merge this now, but I also don't mind if it goes in later.

SGSSGene (approved):

PR #182, opened 2022-07-21
Remove serial mode

sarahet:
Remove serial mode, leave current test infrastructure intact but search tests with SIMD only


src/search_options.hpp
```diff
@@ -92,19 +92,9 @@ struct LambdaOptions : public SharedOptions
...
-    };
-    // ExtensionMode   extensionMode = ExtensionMode::AUTO;
-#ifdef SEQAN_SIMD_ENABLED
-    ExtensionMode   extensionMode = ExtensionMode::FULL_SIMD;
-#else
-    ExtensionMode   extensionMode = ExtensionMode::FULL_SERIAL;
-#endif
+    #ifndef SEQAN_SIMD_ENABLED
+    #error "Lambda must be built with at least SSE4 support. Add -march=native to your compiler flags."
+    #endif
```
h-2, line 97: This is not an ideal place for the error. I would place it at the top of search.cpp right after the header-includes.

h-2 (commented):
Otherwise LGTM

h-2 (approved):

PR #183, opened 2022-08-02
[Test] Switch to SeqAn3 test template

sarahet:
The GitHub actions issue is still on-going I think. Current test times locally:

```
    Start 1: test/cli/index_test
1/2 Test #1: test/cli/index_test ..............   Passed    2.04 sec
    Start 2: test/cli/search_test
2/2 Test #2: test/cli/search_test .............   Passed   77.99 sec
```


test/cli/search_test.cpp
```diff
@@ -0,0 +1,454 @@
...
+                                                   "-r", reduction);
+
+        cli_test_result result_search = execute_app("lambda3", search_command,
+                                                    "-i", index_file,
+                                                    "-q", data(query_file),
+                                                    "-t", "2",
+                                                    "--version-to-outputfile", "0",
+                                                    "-o", output_file);
+
+        EXPECT_EQ(result_search.exit_code, 0);
```
h-2, line 32: This is the place where the test hangs on my computer. I would change it:
```suggestion
        ASSERT_EQ(result_search.exit_code, 0);
```
as it does not make sense to run the subsequent code, if the command invocation already failed. 

This change fixes the hang for me, although all tests still fail.

test/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+
+if (${uppercase_project_name}_HEADER_TEST_ONLY)
+    add_subdirectory (header)
+else ()
+    # Fetch data and add the tests.
+    include (data/datasources.cmake)
+    add_subdirectory (cli)
+endif ()
+
+message (STATUS "${FontBold}You can run `make test` to build and run tests.${FontReset}")
```
h-2, line 85:
```suggestion
message (STATUS "${FontBold}You can run `make cli-test` to build the tests and `make test` to run them.${FontReset}")
```

test/cli/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+target_use_datasources (index_test FILES db_nucl_bs.fasta.gz)
+target_use_datasources (index_test FILES db_prot.fasta.gz)
+target_use_datasources (index_test FILES db_nucl_bs_bifm.fasta.gz.lba)
+target_use_datasources (index_test FILES db_nucl_bs_fm.fasta.gz.lba)
+target_use_datasources (index_test FILES db_nucl_bifm.fasta.gz.lba)
+target_use_datasources (index_test FILES db_nucl_fm.fasta.gz.lba)
+target_use_datasources (index_test FILES db_prot_bifm.fasta.gz.lba)
+target_use_datasources (index_test FILES db_prot_fm.fasta.gz.lba)
+target_use_datasources (index_test FILES db_trans_bifm.fasta.gz.lba)
+target_use_datasources (index_test FILES db_trans_fm.fasta.gz.lba)
```
h-2, line 14: Maybe put multiple files on one line? Or use linebreaks and write filenames on individual line (without repeating `target_use...`).

test/cli/cli_test.hpp
```diff
@@ -0,0 +1,92 @@
...
+    template <typename... CommandItemTypes>
+    cli_test_result execute_app(CommandItemTypes &&... command_items)
+    {
+        cli_test_result result{};
+
+        // Assemble the command string and disable version check.
+        std::ostringstream command{};
+        command << "SEQAN3_NO_VERSION_CHECK=1 " << BINDIR;
+        int a[] = {0, ((void)(command << command_items << ' '), 0) ... };
+        (void) a;
```
h-2, line 39: I think the following should work as well and be more readable:
```suggestion
        ((command << command_items << ' '), ...);
```

h-2: I would actually not use the stringstream at all, and just use `std::string` and ` +=` but I don't care much about this.

test/cli/index_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include "cli_test.hpp"
+
+struct index_test : public cli_test
+{
+    void run_index_test(std::string const & index_command,
+                        std::string const & db_file,
+                        std::string const & index_file,
+                        std::string const & index_type,
+                        std::string const & reduction,
+                        std::string const & control_file)
```
h-2, line 13: It's good style to use string_view in modern C++, but if changing it breaks something, don't bother :)

```suggestion
                        std::string_view const control_file)
```
(also in other places)

test/cli/index_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+    std::string expected
+    {
+        "lambda3 - Lambda, the Local Aligner for Massive Biological DatA.\n"
+        "================================================================\n"
+        "    [OPTIONS] COMMAND [COMMAND-OPTIONS]\n"
+        "    Try -h or --help for more information.\n"
+    };
+    ASSERT_EQ(result.exit_code, 0);
+    ASSERT_EQ(result.out, expected);
+    ASSERT_EQ(result.err, std::string{});
```
h-2, line 54: In these cases, I would use `EXPECT_EQ` and not `ASSERT_EQ` as the subsequent steps are valid (and interesting!) even if one of them fails.

test/cli/search_test.cpp
```diff
@@ -0,0 +1,454 @@
...
+            {
+                system(("bzip2 -d " + output_file).c_str());
+                test_output.open(output_file.substr(0, output_file.length() - 4));
+            }
+            else
+            {
+                test_output.open(output_file);
+            }
+
+            std::ifstream control_output (data(control_file));
```
h-2, line 72:
```suggestion
            std::ifstream control_output{control_file};
```
?

test/data/datasources.cmake
```diff
@@ -0,0 +1,237 @@
...
+                    URL_HASH SHA256=e56d8ba576ad313df4f7bae1cd2193e3c11699749a985ecc1807af65b9815fdc)
+
+## Query input files
+
+declare_datasource (FILE queries_nucl.fasta.gz
+                    URL https://raw.githubusercontent.com/h-2/lambda-testdata/main/input_files/queries_nucl.fasta.gz
+                    URL_HASH SHA256=7a8adcc7ee5d967992a0624b9fb704118223abd112534c2c6456104b30ddad88)
+
+declare_datasource (FILE queries_nucl_bs.fasta.gz
+                    URL https://raw.githubusercontent.com/h-2/lambda-testdata/main/input_files/queries_nucl_bs.fasta.gz
```
h-2, line 60: We are targetting the main branch with these URLs. That's not ideal, because that will change and then previous commits/releases won't build any more.
I suggest we create a `3.0`-tag there now, and then use that in this URL. When the files are changed for a release, we can create a new tag there and update the URLs.

test/data/datasources.cmake
```diff
@@ -0,0 +1,237 @@
...
+
+declare_datasource (FILE output_tblastx_fm.m9
+                    URL https://raw.githubusercontent.com/h-2/lambda-testdata/main/output_files/output_tblastx_fm.m9
+                    URL_HASH SHA256=79981091202983e3d07c26876fc8bd6fe3cd214187db0d91606df17aade32577)
+
+declare_datasource (FILE output_tblastx_fm.sam
+                    URL https://raw.githubusercontent.com/h-2/lambda-testdata/main/output_files/output_tblastx_fm.sam
+                    URL_HASH SHA256=15b08fcc6186c99e8b5680a19708f9aba4f04bf0d5d5becac31ac501eaf6bd4e)
+
+
```
h-2, line 189:
```suggestion
```

test/cli/search_test.cpp
sarahet: We need the data() so that it is clear that this is not located in the test output directory but in the data directory. Otherwise the test will fail as it can't find the control file in the test output directory.

test/data/datasources.cmake
sarahet: Sounds good.

test/cli/index_test.cpp
sarahet: Yes, I only changed it to see the error messages when the test was not running through. Will revert that.

test/cli/CMakeLists.txt
sarahet: :nail_care:

test/cli/index_test.cpp
sarahet: This causes many things to break, so I will not bother as suggested :slightly_smiling_face:

test/data/datasources.cmake
sarahet: We should probably wait though in case the search still changes?

test/cli/cli_test.hpp
```diff
@@ -0,0 +1,91 @@
...
+    // Invoke the app execution. The command line call should be given as separate parameters.
+    template <typename... CommandItemTypes>
+    cli_test_result execute_app(CommandItemTypes &&... command_items)
+    {
+        cli_test_result result{};
+
+        // Assemble the command string and disable version check.
+        std::ostringstream command{};
+        command << "SEQAN3_NO_VERSION_CHECK=1 " << BINDIR;
+        ((command << command_items << ' '), ...);
```
h-2, line 38: Shall I create a PR to the app-template? 😆 🙈

h-2 (approved):

PR #184, opened 2022-08-09
Iterative search

h-2:
~~THIS IS A DRAFT~~

We now have the following extra command line arguments:

```
--search0 [bool] whether to perform the pre-search/iterative-search
--seed-length0
--seed-offset0
--seed-delta0
```

Where the respective seeding-options apply to the pre-search step. By default, everything is deactivated. When activated, length and offset are the same as before, but delta is set to 0. This type of search should never reduce the #queries with one hit, but improves performance by 10-20%.

I have had promising results with the following configs
```
--search0 1 --seed-length0 9 --seed-offset0 9 --seed-delta0 0 --seed-length 11 --seed-offset 11 // basically -1 pre and +1 post
--search0 1 --seed-length0 9 --seed-offset0 3 --seed-delta0 0 --seed-length 12 --seed-offset 4 // longer seed but more offset
--search0 1 --seed-length0 9 --seed-offset0 3 --seed-delta0 0 --seed-length 11 --seed-offset 8 // "best" sensitivity
```
These yield speed-ups of up to 30-70%, but none of these reach previous sensitivity. Results also differ a lot between datasets. For example, the last config only costs 1% sensitivity on the 125 transcriptome dataset, it costs ~4% on the 251 WGS dataset.

### DIFF
--- a/src/search.cpp
+++ b/src/search.cpp
@@ -365,24 +365,27 @@ void realMain(LambdaOptions     const & options)
 
         while (true)
         {
-            localHolder.reset();
-
-            // load records until batch is full or file at end
-            for (auto & [ id, seq ] : file_view)
+            if (localHolder.iterativeSearch != IterativeSearchMode::PHASE2)
             {
-                localHolder.qryIds.push_back(std::move(id));
-                localHolder.qrySeqs.push_back(std::move(seq));
+                localHolder.reset();
 
-                if (++localHolder.queryCount == globalHolder.records_per_batch)
+                // load records until batch is full or file at end
+                for (auto & [ id, seq ] : file_view)
+                {
+                    localHolder.qryIds.push_back(std::move(id));
+                    localHolder.qrySeqs.push_back(std::move(seq));
+
+                    if (++localHolder.queryCount == globalHolder.records_per_batch)
+                        break;
+                }
+
+                if (localHolder.queryCount == 0) // no more records in file
                     break;
+
+                globalHolder.queryCount += localHolder.queryCount; // atomic can be written from multiple threads
             }
 
-            if (localHolder.queryCount == 0) // no more records in file
-                break;
-
-            globalHolder.queryCount += localHolder.queryCount; // atomic can be written from multiple threads
-
-            localHolder.resetViews(); // views reset after sequences have been loaders
+            localHolder.resetViews(); // views reset after sequences have been loaded
 
             // seed
         #ifdef LAMBDA_MICRO_STATS
@@ -403,9 +406,15 @@ void realMain(LambdaOptions     const & options)
                 printProgressBar(lastPercent, curPercent);
             }
 
+            // Check which queries are already successfull
+            iterativeSearchPre(localHolder);
+
             // write to disk
             if (localHolder.blastMatches.size() > 0)
                 writeRecords(localHolder);
+
+            // prepare second phase if necessary
+            iterativeSearchPost(localHolder);
 
         } // implicit thread sync here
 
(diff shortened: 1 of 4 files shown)


src/search_datastructures.hpp
```diff
@@ -475,6 +482,7 @@ class LocalDataHolder
     typename TGlobalHolder::TRedQrySeqs     redQrySeqs = transQrySeqs | redView<TGlobalHolder::c_transAlph,
                                                                                 TGlobalHolder::c_redAlph>;
     size_t                                  queryCount = 0;
+    std::vector<int>                        successfulQueries; // vector<bool> -.-
```
sarahet, line 485: :smile:

src/search.cpp
```diff
@@ -365,24 +365,27 @@ void realMain(LambdaOptions     const & options)
...
-                break;
+                if (localHolder.queryCount == 0) // no more records in file
+                    break;
 
-            globalHolder.queryCount += localHolder.queryCount; // atomic can be written from multiple threads
+                globalHolder.queryCount += localHolder.queryCount; // atomic can be written from multiple threads
+            }
 
-            localHolder.resetViews(); // views reset after sequences have been loaders
+            localHolder.resetViews(); // views reset after sequences have been loaded
```
sarahet, line 388: Why are the views resetted when in Phase 2?

src/search_datastructures.hpp
```diff
@@ -517,9 +528,15 @@ class LocalDataHolder
     LocalDataHolder & operator=(LocalDataHolder &&) = delete;
 
     LocalDataHolder(LambdaOptions const & _options, TGlobalHolder & _gH) :
-        options{_options}, gH{_gH}, stats{},
-        searchScheme{search_schemes::expand(search_schemes::generator::pigeon_opt(0, options.maxSeedDist), options.seedLength)} // fixed search scheme
-    {}
+        options{_options}, gH{_gH}, searchOpts{options.searchOpts}, stats{},
+        searchScheme{search_schemes::expand(search_schemes::generator::pigeon_opt(0, options.searchOpts.maxSeedDist), options.searchOpts.seedLength)} // fixed search scheme
```
sarahet, line 532: This is probably a stupid question (apologies) but are these initialized before the function body is executed?

src/search.cpp
h-2: Because some of the query sequences (the successful ones) have been removed from the set and the views still refer to the old entries.

h-2: (resetting views is a constant time operation, so it also doesn't matter in case no sequences were removed at all)

src/search_datastructures.hpp
h-2: Yep, they are guaranteed to be initialised before.

sarahet: But wouldn't we want the seed length from the search0 options if the search is iterative? Or am I missing something?

h-2:
I previously forgot to add the necessary code for bidirectional search.

sarahet (approved):
Can you just quickly reply to my previous question? I am sure I am missing something but would be good to know what.

src/search_datastructures.hpp
h-2: That is set 3 lines further down in the file. Or what do you mean?

h-2:
But I do have a typo for the bifm scheme initialisation.

src/search_datastructures.hpp
sarahet: Ah nevermind, this anyways changed now. Did not connect it. Now I get it with the updated version.

sarahet:
Then I missed it :eyes:

PR #187, opened 2022-08-17
[feature] minimum bit-score

h-2:
The evalue check is still performed by default right now.


h-2:
Merging this as more PRs are incoming.

PR #188, opened 2022-08-17
Automatic code-formatting

h-2:


h-2:
@sarahet I don't know if you want to look at any of this? If yes, go by commit and skip the "reformat" commit as that just contains whitespace changes (and some include fixes).

Important: In the future, before you push something, run `make reformat` locally. This ensures all your changes conform to the style. `clang_format-14` is required for this.

PR #190, opened 2022-09-21
[fix] view madness in seqan2

### LINKED ISSUE #189: gcc 12 + lambda3 branch == compliation error
https://github.com/seqan/lambda/issues/124#issuecomment-1252447313

h-2:
fixes #189 

---

SeqAn2 deep internally expects things to be default-constructible which some of our view are no longer. It is also no longer possible to workaround this... but it is possible to specialise lots of stuff within in SeqAn2 to not trigger any default-initialisations which is super-unsafe... but works ™️ 

---

SeqAn2 is both 💩 and 🎉 because while the craziness is infinite, you can always work around it, by adding more 💩

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -94,7 +94,7 @@ jobs:
         run: |
           mkdir build
           cd build
-          cmake ../lambda -DLAMBDA_COMPILE_THREADS=2 -DLAMBDA_NATIVE_BUILD=ON -DCMAKE_CXX_FLAGS="${{ matrix.cxx_flags }}" -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache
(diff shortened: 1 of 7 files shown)


h-2:
For some reason, we seem to be loosing gap symbols 🙈

sarahet:
Why do all tests fail? :thinking: And what does it mean for us that we are losing gap symbols?

h-2:
> Why do all tests fail?

Because we are losing gap symbols!

> And what does it mean for us that we are losing gap symbols?

That the tests fail 😜 

I will have a look at it later today! No need to review anything yet

h-2:
ready for review now!

h-2:
1. I know why the GCC+macOS test is failing, but I don't know why it didn't fail before. The problem is that the `mdsum` command is called just `md5` on macOS and FreeBSD. How could this not have failed before and why does it not fail on GCC11?
2. I have no clue why the GCC10 tests on Linux fail. They pass on my GCC on my laptop.

sarahet (approved):

PR #196, opened 2022-11-08
[patch, test] md5 generates a different output to md5sum

SGSSGene:
Output of `md5sum somefile.txt`:
```
a4d518904c07e3e75fa612890e5c813  somefile.txt
```

Output of `md5 somefile.txt`:
```
MD5 (somefile.txt) = a4d518904c07e3e75fa612890e5c813
```

The way the test works, is they take the first value from the file, which is not the hash value in case of `md5`.

Fix is to run `md5 -r somefile.txt` which outputs:
```
a4d518904c07e3e75fa612890e5c813 somefile.txt
```


sarahet (approved):
LGTM, thanks for catching this!

PR #194, opened 2022-10-28
[cli] change defaults, add profiles

h-2:
* adds profiles
* changes default AR to Li10
* changes default search-parameters
* makes min-bit-score default and deactivates max-evalue default

PR #198, opened 2022-11-18
Add protein and nucleotide profiles

sarahet:
This PR includes #194 and #191 and adds nucleotide profiles as well as updated tests matching changes in index  and the search default settings.

### DIFF
--- a/src/search_options.hpp
+++ b/src/search_options.hpp
@@ -91,8 +91,8 @@ struct LambdaOptions : public SharedOptions
     int32_t misMatch      = -3; // only for manual
 
     int32_t  band        = -3;
-    int32_t  minBitScore = -1;
-    double   maxEValue   = 1e-04;
+    int32_t  minBitScore = 42;
+    double   maxEValue   = -1;
     int32_t  idCutOff    = 0;
     uint64_t maxMatches  = 256;
 
@@ -102,7 +102,8 @@ struct LambdaOptions : public SharedOptions
     int32_t preScoring       = 2; // 0 = off, 1 = seed, 2 = region
     double  preScoringThresh = 2.0;
 
-    bool iterativeSearch = false;
+    bool        iterativeSearch = true;
+    std::string profile         = "none";
 };
 
 void parseCommandLine(LambdaOptions & options, int argc, char const ** argv)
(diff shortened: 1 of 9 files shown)


src/search_options.hpp
```diff
@@ -91,8 +91,8 @@ struct LambdaOptions : public SharedOptions
     int32_t misMatch      = -3; // only for manual
 
     int32_t  band        = -3;
-    int32_t  minBitScore = -1;
-    double   maxEValue   = 1e-04;
+    int32_t  minBitScore = 42;
+    double   maxEValue   = -1;
     int32_t  idCutOff    = 0;
     uint64_t maxMatches  = 256;
```
h-2, line 97: I forgot to change this to 25

src/search_options.hpp
```diff
@@ -495,6 +510,42 @@ void parseCommandLine(LambdaOptions & options, int argc, char const ** argv)
         options.geneticCodeQry = static_cast<seqan3::genetic_code>(geneticCodeTmp);
     }
 
+    if (options.profile == "fast")
+    {
+        if (options.nucleotide_mode)
+        {
+            options.iterativeSearch        = false;
```
h-2, line 517: I thought we came to the conclusion that the iterativeSearch is always beneficial, or did we not find a good parameter-pair for nucleotide+fast?

src/search_options.hpp
h-2: This means you need to recreate the search test files I am afraid...

src/search_options.hpp
```diff
@@ -431,51 +442,55 @@ void parseCommandLine(LambdaOptions & options, int argc, char const ** argv)
...
-    parser.add_line("fast (high similarity):    --seed-delta-increases-length on", false);
-    parser.add_line("sensitive (lower similarity): --seed-offset 3", false);
+    // TODO this is only used in serial mode right now
+    // parser.add_option(options.band,
+    //                   'b',
+    //                   "band",
+    //                   "Size of the DP-band used in extension (-3 means log2 of query length;"
+    //                   " -2 means sqrt of query length; -1 means full dp; n means band of size 2n+1)",
+    //                   seqan3::option_spec::advanced,
+    //                   seqan3::arithmetic_range_validator{-3, 1000});
```
h-2, line 468: If nothing uses this, I would propose to remove it (also the option itself). No point in carrying around more legacy codepaths.

h-2:
> Do we want tests for the different profiles? I didn't think it was necessary given that we also don't test all other parameters so I did not add them but let me know if you think differently

I don't think it would hurt to have at least some tests for the other profiles. Even if we don't test all options/combinations, testing the advertised ones gives us a better coverage.

src/search_options.hpp
sarahet: We did not find a good pair and you suggested to pick one of the search0 candidates as regular search for fast.

src/search_options.hpp
sarahet: No problem

src/search_options.hpp
sarahet: I guess depends if we ever want to introduce it again, that's why I left it. But we can remove it, just wanted to check.

sarahet:
> Even if we don't test all options/combinations, testing the advertised ones gives us a better coverage.

Should we test ```m8``` and ```sam``` with unidirectional index for fast and sensitive?

src/search_options.hpp
h-2: If we add it again, we can also add the option back. Considering other people (including future-me), I think it's better to reduce the amount of commented-out code!

h-2:
> Should we test m8 and sam with unidirectional index for fast and sensitive?

Sounds good!

h-2 (approved):

PR #200, opened 2022-12-13
[feature] compressed indexes

h-2:
@sarahet Can you give this a try with a run on a big database? Just a single run should already be helpful.

Also, it would be great to see the filesizes of the compressed VS uncompressed index. For Uniprot, it is roughly 50%.

If also possible, can you check the CPU-Usage during saving and loading the index? It is quite difficult for me to measure on small databases...


sarahet:
Yes, I will do that

src/mkindex.cpp
```diff
@@ -214,23 +216,25 @@ void realMain(LambdaIndexerOptions const & options)
...
-        std::ofstream             os{options.indexFilePath.c_str()};
-        cereal::JSONOutputArchive oarchive(os);
-        oarchive(cereal::make_nvp("lambda index", f));
-    }
-    else
-    {
-        throw 59;
-    }
+        std::string                  filename(options.indexFilePath);
+        bio::io::transparent_ostream os{filename, {.threads = options.threads}};
```
h-2, line 221:
```suggestion
        bio::io::transparent_ostream os{filename, {.threads = options.threads,
                                                   .compression_level = 5}};
```

src/search_algo.hpp
```diff
@@ -63,23 +63,19 @@
...
-    // //     stat(indexFilePath.c_str(), &path_stat);
-    //     if (stat(indexFilePath.c_str(), &path_stat) || !S_ISDIR(path_stat.st_mode))
-    //         throw IndexException("Index directory does not exist or is not readable.\n");
-
     fake_index_file f{options.indexFileOptions};
 
-    if (options.indexFilePath.extension() == ".lba")
+    std::string                  filename(options.indexFilePath);
+    size_t                       threads = std::min<size_t>(options.threads, 4); // more than four threads is harmful
+    bio::io::transparent_istream is{filename, {.threads = threads}};
```
h-2, line 70:
```suggestion
    bio::io::transparent_istream is{filename, {.threads = 2}};
```

Can you try this for decompression?

PR #201, opened 2022-12-21
Re-activate pre-scoring = 1 if no reduction is used

sarahet:
This PR re-activates the pre-scoring default for unreduced sequences as stated in the help message (mainly important for nucleotide mode). Additionally, some minor fixes that have been already adjusted in lambda3 but were missing in lambda2.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "seqan"]
 	path = include/seqan
-	url = git://github.com/seqan/seqan.git
+	url = ../../seqan/seqan.git
--- a/src/shared_misc.hpp
+++ b/src/shared_misc.hpp
@@ -26,7 +26,10 @@
 #include <locale>
 #include <type_traits>
 #include <forward_list>
-#include <sys/sysctl.h>
+
+#if __has_include(<sys/sysctl.h>)
+    #include <sys/sysctl.h>
+#endif
 
 #include <seqan/basic.h>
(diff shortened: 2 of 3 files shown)


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "seqan"]
 	path = include/seqan
-	url = git://github.com/seqan/seqan.git
+	url = https://github.com/seqan/seqan.git
```
h-2, line 3:
```suggestion
	url = ../../seqan/seqan.git
```

While we are at it, we can just switch to relative syntax (this will preserve whatever protocol is used for the main repo).

src/shared_misc.hpp
```diff
@@ -26,7 +26,10 @@
 #include <locale>
 #include <type_traits>
 #include <forward_list>
-#include <sys/sysctl.h>
+
+#if defined(__APPLE__)
```
h-2, line 30:
```suggestion
#if __has_include(<sys/sysctl.h>)
```

This header is also used by FreeBSD and potentially other systems; just not my Linux any more. Can you check if the proposed change works? It might still pick up the header on Linux, but it shouldn't...

h-2 (changes requested):

src/shared_misc.hpp
sarahet: works

h-2 (approved):

PR #203, opened 2023-01-24
[misc] allow dna15 in subject sequences

h-2:
@sarahet this does not add a test. Do you think we need that?


h-2:
The mac tests unbroke themselves 🎉

sarahet:
hmm good question .. maybe would be nicer and maybe a simple solution is just to introduce few characters in one of the existing databases and make it an extra test?

sarahet (approved):

PR #204, opened 2023-03-10
Query preload

h-2:
can you give this a try?

There is a command-line parameter `--lazy-query` which is off by default, i.e. the new behaviour is the default.

### DIFF
--- a/src/search.cpp
+++ b/src/search.cpp
@@ -76,7 +76,7 @@ int searchMain(int const argc, char const ** argv)
     parseCommandLine(options, argc, argv);
 
 #ifdef _OPENMP
-    omp_set_num_threads(options.threads - 1);
+    omp_set_num_threads(options.threads - options.lazyQryFile); // reserve one thread for I/O when lazy-loading
 #else
     options.threads = 1;
 #endif
@@ -337,7 +337,10 @@ void realMain(LambdaOptions const & options)
 
     loadDbIndexFromDisk(globalHolder, options);
 
-    countQuery(globalHolder, options);
+    if (options.lazyQryFile)
+        countQuery(globalHolder, options);
+    else
+        loadQuery(globalHolder, options);
 
     myWriteHeader(globalHolder, options);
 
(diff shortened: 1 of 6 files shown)


h-2:
view fail on GCC < 12. I will fix it on Monday.

src/shared_definitions.hpp
```diff
@@ -254,13 +254,13 @@ template <AlphabetEnum c_origQryAlph, AlphabetEnum c_transAlph, AlphabetEnum c_r
...
 {
     if constexpr (c_redAlph == AlphabetEnum::DNA3BS)
-        return bio::views::add_reverse_complement | views::duplicate;
+        return bio::views::type_reduce | bio::views::add_reverse_complement | views::duplicate;
     else if constexpr (bio::alphabet::nucleotide_alphabet<_alphabetEnumToType<c_redAlph>>)
-        return bio::views::add_reverse_complement;
+        return bio::views::type_reduce | bio::views::add_reverse_complement;
     else if constexpr (c_origQryAlph == c_transAlph)
-        return bio::views::type_reduce;
+        return bio::views::type_reduce | bio::views::type_reduce;
```
sarahet, line 261: Why is this applied twice?

src/search.cpp
```diff
@@ -346,35 +349,55 @@ void realMain(LambdaOptions const & options)
...
+    /* Setup lazy-loader for query (if set in options) */
+    decltype(createQryView(globalHolder, options)) file_view;
+    if (options.lazyQryFile)
+        file_view = createQryView(globalHolder, options);
 
     SEQAN_OMP_PRAGMA(parallel)
     {
         TLocalHolder localHolder(options, globalHolder);
 
+        /* Define region of query this thread get (eager-loading only) */
```
sarahet, line 364: gets

sarahet: Should this be in an if statement because it does not apply to lazy-loading?

h-2: > Should this be in an if statement because it does not apply to lazy-loading?

The variables need to be defined because of scope either way. It only happens once when the thread starts, so it doesn't influence performance.

sarahet (approved):

PR #207, opened 2023-04-20
Add parameters for bisulfite profiles

sarahet:
This PR defines the parameters for the bisulfite default, fast and sensitive profiles.

I realized that we have no recommendations for E-value cut-offs anymore, now that we switched to the default bit score. Should we change that somehow?


sarahet:
I need to still update the test files, will do that soon.

sarahet:
> I need to still update the test files, will do that soon.

Done

h-2 (approved):

PR #212, opened 2023-05-08
Bits2

h-2:
We don't have tests on this branch, but I have done some local tests that all look fine.

To use a minimum bit-score and no e-value, add `--e-value 100 --bit-score 42`.

@sarahet Can you run the following on a uniref DB and one small query file (maybe just 5 or 10MB):

1) lambda2 master branch with e-value 1e-4, one thread
2) this branch with e-value 1-e4, one thread
3) this branch with bit-score 60, one thread

The output of 1) and 2) should be identical. When filtering 1) for bitscore 60, the output should be very similar to 3 (ideally identical).
(This doesn't have to be on our benchmark server)

### DIFF
--- a/src/search_options.hpp
+++ b/src/search_options.hpp
@@ -95,7 +95,8 @@ struct LambdaOptions : public SharedOptions
     int             misMatch        = 0; // only for manual
 
     int             xDropOff    = 0;
-    int32_t         minBitScore = -1;
+    int             band        = -1;
+    double          minBitScore = 0;
     double          maxEValue   = 1e-04;
     int             idCutOff    = 0;
     unsigned long   maxMatches  = 500;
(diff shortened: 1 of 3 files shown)


src/search_options.hpp
```diff
@@ -247,6 +248,14 @@ parseCommandLine(LambdaOptions & options, int argc, char const ** argv)
     setMinValue(parser, "e-value", "0");
     setMaxValue(parser, "e-value", "100");
 
+    addOption(parser, ArgParseOption(" ", "bit-score",
```
sarahet, line 251: This needs to be an empty string

sarahet:
> The output of 1) and 2) should be identical. When filtering 1) for bitscore 60, the output should be very similar to 3 (ideally identical).

1 and 2 are identical, 1 filtered for bitscore 60 is identical to 3 (one query, 10Mb with uniref)

sarahet:
> To use a minimum bit-score and no e-value, add --e-value 100 --bit-score 42.

This is different to Lambda3, is that because otherwise it would be difficult to merge with Lambda2's code?

h-2:
> > To use a minimum bit-score and no e-value, add --e-value 100 --bit-score 42.
> 
> This is different to Lambda3, is that because otherwise it would be difficult to merge with Lambda2's code?

Yes, I would have had to change more code, and I wanted to keep all changes to lambda2 minimal.

sarahet (approved):
LGTM

PR #215, opened 2023-07-11
Switch to max evalue

sarahet:


src/search_options.hpp
```diff
@@ -93,8 +93,8 @@ struct LambdaOptions : public SharedOptions
     int32_t match         = 2;  // only for manual
     int32_t misMatch      = -3; // only for manual
 
-    int32_t  minBitScore = 42;
-    double   maxEValue   = -1;
+    int32_t  minBitScore = -1;
+    double   maxEValue   = 0.01;
```
h-2:
```suggestion
    double   maxEValue   = 1-e2;
```

h-2: Prefer scientific notations for floats smaller than one

src/search_options.hpp
```diff
@@ -263,6 +258,11 @@ void parseCommandLine(LambdaOptions & options, int argc, char const ** argv)
                         .validator = sharg::arithmetic_range_validator{-1, 100}
     });
 
+    if (options.domain == domain_t::bisulfite)
+    {
+        options.maxEValue = 0.000000001;
```
h-2:
```suggestion
        options.maxEValue = 1e-9;
```

h-2 (commented):
I think the scientific notation is a lot easier to read, especially for the second value.

h-2 (approved):

PR #222, opened 2023-11-17
[fix] alignment bug

h-2:
tests still need fixing

### DIFF
--- a/src/search_algo.hpp
+++ b/src/search_algo.hpp
@@ -911,26 +911,23 @@ inline void _writeRecord(TBlastRecord & record, TLocalHolder & lH)
 template <typename TBlastMatch, typename TLocalHolder>
 inline void _setupAlignInfix(TBlastMatch & bm, typename TLocalHolder::TMatch const & m, TLocalHolder & lH)
 {
-    int64_t startMod = (int64_t)m.subjStart - (int64_t)m.qryStart;
+    /* create bm coordinates from m coordinates */
 
-    bm.qEnd                = lH.transQrySeqs[m.qryId].size();
-    decltype(bm.qEnd) band = _bandSize(bm.qEnd);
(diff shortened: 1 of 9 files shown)


src/search_misc.hpp
```diff
@@ -46,7 +46,7 @@ struct QueryException : public std::runtime_error
 inline int _bandSize(uint64_t const seqLength)
```
sarahet: Should the return type here also be ```int64_t```?

src/search_algo.hpp
```diff
@@ -1133,7 +1134,48 @@ inline void _performAlignment(TDepSetH &      depSetH,
...
+            {
+                l.subjEnd   = r.subjEnd;
+                r.subjStart = l.subjStart;
+            }
+        }
+
+        // pairwise "swallow" from right to left
+        for (auto it = matches.rbegin(); it < matches.rend() - 1; ++it)
+        {
+            Match & l = *it;
```
sarahet: Given that the comment says from right to left, I understood it in a way that we are still "looking" at the vector the same way as before. However, now that we are backwards iterating technically ```l``` is to the right of ```r``` in the sense of that comment and it confused me naming-wise when I tried to understand what is happening (because it does not match the ```l``` and ```r``` in the loop before order-wise). Can we find a different way of naming this that is more intuitive?

src/search_misc.hpp
h-2: Done. I don't think this affects anything though.

src/search_algo.hpp
h-2: That's a good point. I have switched `r` and `l`. Is that OK?

sarahet: yes, thanks!

sarahet (approved):
LGTM

PR #227, opened 2024-01-08
[fix] check presence of taxtree in index

h-2:


src/mkindex_algo.hpp
```diff
@@ -580,15 +580,19 @@ auto parseAndStoreTaxTree(std::vector<bool> & taxIdIsPresent, LambdaIndexerOptio
...
     {
         if (taxIdIsPresentOrParent[i] && empty(taxonNames[i]))
         {
             std::cerr << "Warning: Taxon with ID " << i << " has no name associated, defaulting to \"n/a\".\n";
             taxonNames[i] = "n/a";
+            ++taxaWithoutNameCount;
         }
     }
+    if (taxaWithoutNameCount * 100 / taxonNames.size() > 10)
+        std::cerr << "Warning: More than 10% of taxa have no valid name entry.\n";
```
SGSSGene: This might should be:
```suggestion
    if (taxaWithoutNameCount * 10 > taxonNames.size())
        std::cerr << "Warning: More than 10% of taxa have no valid name entry.\n";
```
The issue is that `taxaWithoutNameCount * 100 / taxonNames.size()` might be `10.9` which will be rounded to `10` and then no warning will be given even though it's above 10%.

h-2: Yeah, I mean the whole check / the exact value is not super-important, but you are right of course!

sarahet:
LGTM (minus Github mac test issues)

sarahet (approved):

PR #230, opened 2024-07-21
Update README.rst

h-2:
include new paper


sarahet (approved):
LGTM. Some checks are still queuing but probably we can just merge this?

